Issue #7, opened 2015-11-11
remove usage of `sed -E`


zmanji:
The script currently contains:

```
  pants_requirement="pantsbuild.pants"
  pants_version=$(
    grep -E "^[[:space:]]*pants_version" pants.ini 2>/dev/null | \
      sed -E 's|^[[:space:]]*pants_version[:=][[:space:]]*([^[:space:]]+)|\1|'
  )
```

CentOS 5 ships with:

```
$ sed --version
GNU sed version 4.1.5
Copyright (C) 2003 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE,
to the extent permitted by law.
```

which does not support the `-E` parameter:

```
$ echo a | sed -E "s|a|b|"
sed: invalid option -- E
....
```

The use of `-E` should be removed so this script can work on systems that ship with older versions of coreutils.

jsirois:
@Yasumoto expressed interest so I'll assign over to him once he accepts the 'Contributors' invite.

PR #8, opened 2015-11-11
Migrate from sed extended regular expressions

Yasumoto:
GNU sed and BSD sed are only recently converged when it comes to supporting -E.

Older versions of GNU sed use -r to use extended regular expressions.


Yasumoto:
This will close #7

jsirois:
Patched in @ https://github.com/pantsbuild/setup/commit/f33ec0171170fd561f68bc4b7457dae796cfbb7a

Thanks @Yasumoto !

PR #11, opened 2016-02-13
Use different bootstrap dirs per platform

benley:
This makes pants work when a user has their home directory NFS mounted
to multiple OSes.


jsirois:
Thanks @benley! Merged at https://github.com/pantsbuild/setup/commit/990b364cd196cdeee3ac9ec9bc4607df444bbb9f

PR #13, opened 2016-04-21
Upgrade default virtualenv version

ewang:
Upgrade default virtualenv version to 15.0.0 in the setup script.

As far as I can tell, the only breaking change is virtualenv dropping support for Python 3.2 in 14.0.0.

This resolves warning messages about using older versions of pip, and also fixes some edge cases where pip < 8 doesn't install certain packages.


jsirois:
A belated thanks for this. Virtualenv has since been upgraded past 15.0.0 so I'm going to close.

Issue #14, opened 2016-10-26
Unable to install pants using setup due to missing setuptools


derektamsen:
When running `curl -L -O https://pantsbuild.github.io/setup/pants && chmod +x pants && touch pants.ini` I am seeing the following issue in the [openjdk docker container](https://hub.docker.com/_/openjdk/) based on Debian Jessie:

``` shell
# ./pants -V
New python executable in /root/.cache/pants/setup/bootstrap-Linux-x86_64/pants.Y6bv1k/install/bin/python2.7
Also creating executable in /root/.cache/pants/setup/bootstrap-Linux-x86_64/pants.Y6bv1k/install/bin/python
Installing pip, wheel...done.
Collecting pantsbuild.pants
  Using cached pantsbuild.pants-1.1.0.tar.gz
Could not import setuptools which is required to install from a source distribution.
Please install setuptools.
./pants: line 84: /root/.cache/pants/setup/bootstrap-Linux-x86_64/unspecified/bin/python: No such file or directory
```

I am able to resolve this issue by reversing the change made in commit a4b5f74 and installing setuptools in the venv

jsirois:
Reproed. I'm going to spend 5 or 10 minutes trying to find a fix that covers both this case and https://github.com/pantsbuild/pants/issues/3948 properly.

jsirois:
OK - I think this is the right short-term solution, basically emulate the pex setuptools constraint in the `pantsbuild.pants` install to force both setuptools to be installed, but also installed pinned to a low version in the cases like that seen in https://github.com/pantsbuild/pants/issues/3948:

```
jsirois@gill ~/dev/pantsbuild/setup (gh-pages *) $ git diff
diff --git a/pants b/pants
index f9089e7..c87ac44 100755
--- a/pants
+++ b/pants
@@ -20,6 +20,11 @@ PANTS_BOOTSTRAP="${PANTS_HOME}/bootstrap-$(uname -s)-$(uname -m)"

 VENV_VERSION=15.0.2

+# This requirement is installed alongside pantsbuild.pants to hack around
+# interpreters that have newer setuptools already installed, effectively
+# re-installing an older setuptools and pinning low.
+SETUPTOOLS_REQUIREMENT="setuptools>=2.2,<20.11"
+
 VENV_PACKAGE=virtualenv-${VENV_VERSION}
 VENV_TARBALL=${VENV_PACKAGE}.tar.gz

@@ -72,8 +77,8 @@ function bootstrap_pants {
[...]

PR #15, opened 2016-10-26
Work around pants/pex old setuptools req.

jsirois:
This addresses #14.

### DIFF
--- a/pants
+++ b/pants
@@ -23,10 +23,10 @@ VENV_VERSION=15.0.2
 # This requirement is installed alongside pantsbuild.pants to hack around
 # interpreters that have newer setuptools already installed, effectively
 # re-installing an older setuptools and pinning low to the version range
-# known to be ingestable by pants and pex. See:
+# known to be ingestable by both pants and pex. See:
 #   https://github.com/pantsbuild/pants/issues/3948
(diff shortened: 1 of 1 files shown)


jsirois:
@kwlzn or @benjyw can you take a look and check this seems sane?

pants
```diff
@@ -20,6 +20,14 @@ PANTS_BOOTSTRAP="${PANTS_HOME}/bootstrap-$(uname -s)-$(uname -m)"
 
 VENV_VERSION=15.0.2
 
+# This requirement is installed alongside pantsbuild.pants to hack around
+# interpreters that have newer setuptools already installed, effectively
+# re-installing an older setuptools and pinning low to the version range
+# known to be ingestable by pants and pex. See:
+#   https://github.com/pantsbuild/pants/issues/3948
+#   https://github.com/pantsbuild/setup/issues/14
+SETUPTOOLS_REQUIREMENT="setuptools>=2.2,<20.11"
```
kwlzn: should the lower range of this pin be `>=5.4.1` given that's whats in our requirements.txt?

kwlzn (approved):
lgtm, one thought

pants
jsirois: Yeah - sgtm

PR #29, opened 2018-12-06
Show that 1.13.0.dev0 is broken _publicly_.

jsirois:


jsirois:
This was for demonstraion purposes re detecting the libssl bug publically. I'm closing, but have filed https://github.com/pantsbuild/pants/issues/6875 to track doing this automatically.

Issue #30, opened 2019-03-11
Add support for Python 3


Eric-Arellano:
With Pants soon to be released with Python 3 wheels https://github.com/pantsbuild/pants/pull/7197 and the Python 3 migration nearing completion https://github.com/pantsbuild/pants/issues/6062, we need to modify this setup script to allow Pants end users to consume the upgrade.

## Goals for an ideal solution
1. Decision of which interpreter to use applies to the whole organization. 
   * Typically, an individual or team will be responsible for upgrading Pants. They should be the only ones to have to know this is even a decision. Every day users should not be concerned with this upgrade.
   * Thus, solution must be tracked by VCS.
   * Solution should not require something like each individual modifying their `.bashrc` with their individual preference. It should be an organization-level decision set for the whole repo.
1. Transparent what is being used.
    * We certainly shouldn't print out to the console every time or anything. But we should make it possible to find, e.g. in a config file.
   * Will help with debugging.
1. Possible to change the interpreter version.
   * For example, deciding to upgrade from 2.7 -> 3.6 or 3.6 -> 3.7.
   * This decision isn't expected to be very frequent, but it will sometimes happen.
   * When a change happens, must ensure that no one has to do anything beyond the admin. For example, we can't expect people to have to `rm -rf` their cache.
   * Non-goal: make it extremely easy to change back-and-forth between interpreter versions, a la `./pants2` and `./pants`. We are not expecting the average Pants user to worry about which Python interpreter they use to run Pants as both should work equally as well and it is more cognitive load than necessary.
1. Simple to make the first-time decision.
   * Clearly explain what the options are.
   * Don't overwhelm with too intricate of details.
   * Emphasize Py3 will soon be required for runtime.
   * Maybe interactive?
1. Backwards compatible.
   * If the user doesn't change their behavior at all, Pants should continue working as before, i.e. continue using Python 2.7.
   * Note this doesn't mean that we can't use the new venv naming scheme. That's okay. Just Pants needs to keep working when the user doesn't change their behavior from before.

## Proposed solution
Users should add `pants_runtime_python_version: {2.7,3.6,3.7}` to their `pants.ini` `[GLOBAL]` entry, just as they set `pants_version`. 

We would update the website at https://www.pantsbuild.org/install.html to provide instructions, along the lines of:

```
Pants may be ran with Python 2.7, 3.6, or 3.7. This does not mean your code has to use this Python
version, as Pants will spawn subprocesses to run with the interpreter you want to use for your own
[...]

PR #32, opened 2019-03-12
Add support for running with Python 3.6 and 3.7

Eric-Arellano:
### Problem
Now that we're releasing Pants with Python 3 wheels (https://github.com/pantsbuild/pants/pull/7197, overall migration tracked by https://github.com/pantsbuild/pants/issues/6062), we need to add a way for every day Pants users to consume the upgrade.

See https://github.com/pantsbuild/setup/issues/30 for more context, including the design goals.

### Solution
Parse the `pants_runtime_python_version` from `pants.ini` to determine which Python interpreter to use. If it's missing, default to Python 2.7.

This requires modifying the venv folders to support multiple Python versions. Rather than using the naming scheme `${pants_version}`, we now consistently use `${pants_version}_py{py_major_minor}`, regardless of if the user is using `pants_runtime_python_version`.

It also requires upgrading virtualenv to avoid a deprecation warning when running with Python 3.

Finally, we add tests to ensure this all works.

### Result
Running `./pants` like normal for the first time will bootstrap Pants again and create the new folder `1.15.dev3_py{27,36,37}`.

Changing the `$PYTHON` env var or the `pants_runtime_python_version` in `pants.ini` will change which Python version is used to run Pants.

When changing interpreters for the first time, it will bootstrap Pants again. After that, you can change without issue.


Eric-Arellano:
Reviewers: CI is going to fail on this one until we release `1.15.0.dev4`, because we need the Py3 wheels and https://github.com/pantsbuild/pants/pull/7363 to parse `pants_engine_python_version` in `pants.ini`.

Stu and I came up with a strategy to (hopefully) still land this change before sending the release email to pants-devel. Right after we finish the release to PyPi, I'll update the `pants.ini` to point to the new version and run CI. If all goes green, we'll merge this PR and send the email to pants-devel about how to consume this upgrade. If CI fails, we simply push off the announced release one week and go with a silent release.

ci.py
```diff
@@ -68,5 +87,37 @@ def setup_pants_version(test_pants_version: PantsVersion):
     f.writelines(original_pants_ini)
 
 
+@contextmanager
+def setup_python_version(test_python_version: PythonVersion):
+  """Modify pants.ini to allow the Python version to be unspecified or change to what was requested."""
+  expected_prefix = "pants_engine_python_version"
```
stuhood: This was renamed in https://github.com/pantsbuild/pants/pull/7363 I think. Ditto the PR description, and a few other locations in here.

ci.py
```diff
@@ -68,5 +87,37 @@ def setup_pants_version(test_pants_version: PantsVersion):
     f.writelines(original_pants_ini)
 
 
+@contextmanager
+def setup_python_version(test_python_version: PythonVersion):
```
stuhood, line 97: This is a pretty large method... might be good to extract some helpers.

pants
```diff
@@ -13,11 +13,15 @@
...
-PYTHON=${PYTHON:-$(which python2.7)}
-
 PANTS_HOME="${PANTS_HOME:-${XDG_CACHE_HOME:-$HOME/.cache}/pants/setup}"
 PANTS_BOOTSTRAP="${PANTS_HOME}/bootstrap-$(uname -s)-$(uname -m)"
 
+interpreter_version=$(
+  grep -E "^[[:space:]]*pants_engine_python_version" pants.ini 2>/dev/null | \
+    cut -f2 -d: | tr -d " "
+) || interpreter_version=2.7
+PYTHON=${PYTHON:-$(which python${interpreter_version})}
```
stuhood: Is it worth checking that this resulted in a non-empty value, and erroring if not? Or does the `venv` creation result in a useful error message if your choice isn't installed?

stuhood (approved):
Thanks. One blocker related to the option name, but the rest looks reasonable.

ci.py
Eric-Arellano: Good idea! I refactored it. Let me know if it could use another refactor.

pants
```diff
@@ -13,11 +13,21 @@
 
 set -eou pipefail
 
-PYTHON=${PYTHON:-$(which python2.7)}
-
 PANTS_HOME="${PANTS_HOME:-${XDG_CACHE_HOME:-$HOME/.cache}/pants/setup}"
 PANTS_BOOTSTRAP="${PANTS_HOME}/bootstrap-$(uname -s)-$(uname -m)"
 
+interpreter_version=$(
+  grep -E "^[[:space:]]*pants_runtime_python_version" pants.ini 2>/dev/null | \
```
cosmicexplorer: I still definitely think this should:
1. be its own function so you can call it for either `pants_version` or `pants_runtime_python_version`
2. be converted into `sed -rne '/\s*pants_version: / s#^.*: ##p'` as per https://github.com/pantsbuild/setup/issues/30#issuecomment-471805216, as in:
```bash
function get_pants_ini_config_line() {
  config_key="$1"
  sed -rne "/\s*${config_key}: / s#^.*: ##p" pants.ini 2>/dev/null
}
```
If that's too terse, comments from https://github.com/pantsbuild/setup/issues/30#issuecomment-471805216 could be added. I would at least like this to be made into its own function, but using `sed` here gives us more flexibility later if we need it without adding more complexity.

pants
```diff
@@ -77,7 +87,10 @@ function bootstrap_pants {
     pants_version="unspecified"
   fi
 
-  if [[ ! -d "${PANTS_BOOTSTRAP}/${pants_version}" ]]
+  python_major_minor_version=$(${PYTHON} -c 'import sys; print("".join(map(str, sys.version_info[0:2])))')
```
cosmicexplorer: I like the idea of making this into a function so it's less hidden that we're invoking a python subprocess to do something nontrivial here. It also makes it easier to review later. I would also like it if this function used a heredoc, e.g.:
```bash
function python_major_minor_version() {
  python_exe="$1"
  "$python_exe" <<EOF
import sys
major_minor_version = ''.join(str(version_num) for version_num in sys.version_info[0:2])
print(major_minor_version)
EOF
}
```

That's intentionally verbose to highlight how you have more room to work with when making it into a separate function.

ci.py
```diff
@@ -35,37 +46,83 @@ def create_parser() -> argparse.ArgumentParser:
...
+    return list(f.readlines())
+
+
+def write_pants_ini(lines: Iterable[str]) -> None:
   with open(PANTS_INI, 'w') as f:
-    f.writelines(original_pants_ini)
+    f.writelines(lines)
+
+
+def write_config_entry(*, entry_name: str, entry_value: Optional[str]) -> None:
```
cosmicexplorer: Is there a reason this doesn't use [`configparser`](https://docs.python.org/3.7/library/configparser.html)?

cosmicexplorer (changes requested):
This could work, but if this is going to be the script that users run to setup pants, I would like some of the bash logic to be extracted into functions, as that also makes any weird errors more debuggable for the pants user. The `configparser` module seems like it might be able to abbreviate a lot of `ci.py`, but I'm less familiar with that.

pants
```diff
@@ -45,12 +44,41 @@ function tempdir {
...
+  python_exe="$1"
+  "$python_exe" <<EOF
+import sys
+major_minor_version = ''.join(str(version_num) for version_num in sys.version_info[0:2])
+print(major_minor_version)
+EOF
+}
+
+function determine_pants_runtime_python_version {
+  interpreter_version="$(get_pants_ini_config_value pants_runtime_python_version)"
```
```suggestion
  interpreter_version="$(get_pants_ini_config_value 'pants_runtime_python_version')"
```
I tend to use the convention of surrounding string arguments with single quotes, especially when nested in a process output substitution, to differentiate them from function calls.

pants
```diff
@@ -45,12 +44,41 @@ function tempdir {
...
+  if [[ -z "${interpreter_version}" ]]; then
+    interpreter_version='2.7'
+  fi
+  python_bin_name=${PYTHON:-python${interpreter_version}}
```
```suggestion
  python_bin_name="${PYTHON:-python"${interpreter_version}"}"
```
Would definitely have double quotes around all variable dereferences, spaces can cause weird errors and consistency is key (imho). There's no reason not to have them on nested `${}` expressions too.

pants
```diff
@@ -45,12 +44,41 @@ function tempdir {
...
+  if which "${python_bin_name}" >/dev/null; then
+    echo `which ${python_bin_name}`
```
cosmicexplorer: Backtick substitution is "deprecated": http://mywiki.wooledge.org/BashFAQ/082. Many people cargo cult backticking from other scripts, especially for `which` calls, so you'll still see this often, but it should be converted to `"$()"`.

pants
```diff
@@ -45,12 +44,41 @@ function tempdir {
...
+  else
+    echo "${python_bin_name} could not be found on your PATH! Pants will not work." >&2
```
cosmicexplorer: I don't super care about the distinction personally, but I would use `1>&2` here for consistency with the rest of the file.

pants
```diff
@@ -45,12 +44,41 @@ function tempdir {
...
+  fi
+}
+
 # TODO(John Sirois): GC race loser tmp dirs leftover from bootstrap_XXX
 # functions.  Any tmp dir w/o a symlink pointing to it can go.
 
 function bootstrap_venv {
-  if [[ ! -d "${PANTS_BOOTSTRAP}/${VENV_PACKAGE}" ]]
-  then
+  if [[ ! -d "${PANTS_BOOTSTRAP}/${VENV_PACKAGE}" ]]; then
```
cosmicexplorer: Good change!

pants.ini
```diff
@@ -1,6 +1,6 @@
 [GLOBAL]
 pants_version: 1.15.0.dev3
-
 plugins: [
-    'pantsbuild.pants.contrib.go==%(pants_version)s',
-  ]
+	'pantsbuild.pants.contrib.go==%(pants_version)s',
```
cosmicexplorer, line 5: This changing is because of the use of `configparser`, right? This seems like an ok change then, since it seems like that module doesn't provide any options for spacing when writing out files.

pants
```diff
@@ -65,40 +93,40 @@ function bootstrap_venv {
...
+  pants_version="$(get_pants_ini_config_value pants_version)"
+  if [[ -n "${pants_version}" ]]; then
     pants_requirement="${pants_requirement}==${pants_version}"
   else
     pants_version="unspecified"
   fi
 
-  if [[ ! -d "${PANTS_BOOTSTRAP}/${pants_version}" ]]
-  then
+  python_major_minor_version="$(get_python_major_minor_version ${python})"
```
```suggestion
  python_major_minor_version="$(get_python_major_minor_version "${python}")"
```
I would recommend doing this.

pants
```diff
@@ -65,40 +93,40 @@ function bootstrap_venv {
...
 function bootstrap_pants {
+  python="$1"
   pants_requirement="pantsbuild.pants"
-  pants_version=$(
-    grep -E "^[[:space:]]*pants_version" pants.ini 2>/dev/null | \
-      cut -f2 -d: | tr -d " "
-  )
-  if [[ -n "${pants_version}" ]]
-  then
+  pants_version="$(get_pants_ini_config_value pants_version)"
```
```suggestion
  pants_version="$(get_pants_ini_config_value 'pants_version')"
```
I would recommend doing this.

Eric-Arellano:
Reviewers: I broke out https://github.com/pantsbuild/setup/pull/36, https://github.com/pantsbuild/setup/pull/37, and https://github.com/pantsbuild/setup/pull/35 as pre-work for this PR, which are all merged now into this PR for a smaller diff.

This should be easier to review now and is all ready to go. I'm hoping to have it okayed before our release tomorrow, so that we can simply run the tests upon the release, prove it works via CI, and merge. Thanks!

pants
```diff
@@ -62,8 +62,25 @@ function get_pants_ini_config_value {
...
+import sys
+major_minor_version = ''.join(str(version_num) for version_num in sys.version_info[0:2])
+print(major_minor_version)
+EOF
+}
+
+function check_interpreter_on_path {
+  python="$1"
+  if ! which "${python}" >/dev/null; then
+    die "Could not find ${python_bin_name}. Please ensure ${python_bin_name} is on your PATH."
```
stuhood: It's implicit that `python_bin_name` will be set by this point... would probably be preferable to use the parameter or explicitly pass a human readable form if `$1` isn't readable.

stuhood (approved):
Looks good: thanks!

pants
Eric-Arellano: This was 100% a bad merge! 

We intentionally want to keep this function generic, because in the future when we (maybe) auto-generate `pants.ini` I'm envisioning our flow will look something like this: `try py36 -> try py27 -> try py37 -> fail`.

This function—with modifications—will be the basis for that flow.

Thanks!

ci.py
```diff
@@ -74,6 +93,25 @@ def setup_pants_version(test_pants_version: PantsVersion):
...
+  updated_config = read_config()
+  config_entry = "pants_runtime_python_version"
+  if test_python_version == PythonVersion.unspecified:
+    updated_config.remove_option(GLOBAL_SECTION, config_entry)
+  else:
+    updated_config[GLOBAL_SECTION][config_entry] = test_python_version.value
+  write_config(updated_config)
+  try:
+    yield
+  except subprocess.CalledProcessError:
```
illicitonion: I'm not sure why this `except` is here; isn't the behaviour the same without it?

illicitonion (approved):
Works for me, though I don't know really anything about the setup repo :)

pants
```diff
@@ -72,11 +89,17 @@ function get_pants_ini_config_value {
...
   else
-    die "Could not find ${PYTHON_BIN_NAME}. Please ensure ${PYTHON_BIN_NAME} is on your PATH."
+    interpreter_version="$(get_pants_ini_config_value 'pants_runtime_python_version')"
+    if [[ -z "${interpreter_version}" ]]; then
+      interpreter_version='2.7'
+    fi
+    python_bin_name="python${interpreter_version}"
   fi
+  check_interpreter_on_path "${python_bin_name}"
+  which "${python_bin_name}"
```
cosmicexplorer: If you remove the `>/dev/null` in `check_interpreter_on_path` and rename it to `get_exe_path_or_die` (which I would recommend anyway, as it doesn't do anything python-specific), then you can remove this `which`. Checking that an exe exists with `which`, but then returning its path in another `which` call feels error-prone to me (what if we slightly change this method later and we end up checking that one interpreter exists, but returning a path to another one?). The suggestion isn't about code golfing -- checking that something exists and then return a path to it separately feels error-prone. I also find the result (just calling `get_exe_path_or_die "${python_bin_name}"` at the end, which will output the exe path to stdout on success) extremely readable -- I don't know that the `check_interpreter_on_path` function could potentially exit from looking at its callsites.

ci.py
```diff
@@ -22,9 +22,19 @@ def __str__(self):
       return self.value
 
 
+class PythonVersion(Enum):
```
cosmicexplorer, line 25: Love enums!!

cosmicexplorer (approved):
Thanks!!

PR #33, opened 2019-03-12
Add ci.py helper script to allow parametrizing the pants_version

Eric-Arellano:
### Problem
The `./pants` script claims to support running with both a pinned `pants_version` in the `pants.ini` and with the `pants_version` unspecified.

We currently do not test this unspecified branch, which we should be doing to guarantee that functionality always works.

Further, this is pre-work for https://github.com/pantsbuild/setup/pull/32, which will add `--python-version` as another parameter for `ci.py` to allow an unspecified Python version vs. `pants.ini` specifying 2.7 or 3.6 (FYI resulting [`.travis.yml`](https://github.com/pantsbuild/setup/pull/32/files#diff-354f30a63fb0907d4ad57269548329e3)).

### Solution
Introduce a new `ci.py` helper script that uses argparse to have the parameter `--pants-version {unspecified,config}`.

#### Requires adding Python 3 to CI shards
Because this script uses modern Python 3—i.e. enums, f-strings, type hints, keyword-only args, and `subprocess.run()`—it requires Python 3.5+ to run.

Even though we can't yet run `./pants` with Python 3 until https://github.com/pantsbuild/setup/pull/32 lands, this PR does some of the pre-work in adding Python 3 to each of the shards so that we can run this script. This is valid to do because every single shard is going to run the tests with both Py27 and Py36—unlike Pants where we have dedicated shards for each interpreter—so there is only benefit in ensuring each shard has Py27 and Py36.

#### Also updates test's pinned `pants_version`

### DIFF
--- a/ci.py
+++ b/ci.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python3
+# Copyright 2019 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+import argparse
+import os
+import subprocess
+from contextlib import contextmanager
+from enum import Enum
+
+
+class PantsVersion(Enum):
+  unspecified = "unspecified"
+  pants_ini = "pants.ini"
+
+  def __str__(self):
+      return self.value
+
+
+class PythonVersion(Enum):
+  unspecified = "unspecified"
+  py27 = "2.7"
+  py36 = "3.6"
+  py37 = "3.7"
+  
+  def __str__(self):
+    return self.value
+
+
+PANTS_INI = 'pants.ini'
+
+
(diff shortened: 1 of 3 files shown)


ci.py
```diff
@@ -0,0 +1,72 @@
...
+from contextlib import contextmanager
+from enum import Enum
+
+
+PANTS_INI = 'pants.ini'
+
+
+class PantsVersion(Enum):
+  unspecified = "unspecified"
+  pants_ini = "pants.ini"
```
stuhood: Slightly confusing to have an enum choice that looks like a filename. Maybe just `config`?

stuhood (approved):
Looks good. Thanks!

benjyw (approved):

PR #34, opened 2019-03-13
Hotfix for #33 to propagate subprocess return code and consistently use Py36 in Linux

Eric-Arellano:
https://github.com/pantsbuild/setup/pull/33 does not propagate return codes for subprocesses, meaning if the `./pants foo` commands fail, the error will print out to the screen but Travis will show everything as working. Fix this by adding `.check_returncode()`.

Also we found Linux Trusty and Xenial were using Python 3.5 to execute `ci.py`, as they err when introducing `f-strings`, which are 3.6+. So, use `pyenv global` to consistently use the correct interpreters.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,6 +56,8 @@ matrix:
       <<: *linux_setup
       dist: precise
       sudo: required
+      before_install:
+        - pyenv versions
 
     - name: "Ubuntu 14.04 - Trusty"
       <<: *linux_setup
@@ -68,3 +70,5 @@ matrix:
     - name: "Ubuntu 16.04 - Xenial"
       <<: *linux_setup
       dist: xenial
+      before_install:
+        - pyenv versions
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -16,19 +16,24 @@ osx_setup: &osx_setup
...
       LDFLAGS="-L/usr/local/opt/openssl/lib"
       CPPFLAGS="-I/usr/local/opt/openssl/include"
   before_install:
-    - >
+    - &pyenv_install >
       git clone https://github.com/pyenv/pyenv ${PYENV_ROOT}
       && ${PYENV_ROOT}/bin/pyenv install 3.6.8
       && ${PYENV_ROOT}/bin/pyenv global 3.6.8
+  before_script:
+    # Override file handler and thread limits
```
stuhood, line 34: Are we seeing a failure without these? Should we set them consistently across platforms?

stuhood (approved):
Thanks!

.travis.yml
Eric-Arellano: Yes, master sometimes would fail with the `too many open files error`: https://travis-ci.org/pantsbuild/setup/jobs/505521244#L283

However, we didn't catch this because of the missing `check=True` in `subprocess.run()`, meaning the subprocess would fail but not cause the entire `ci.py` script to fail like we would want it to.

--

No idea if we should add it to Linux shards too. Note we do not do this in Pants CI for Linux; we override for every OSX shard and for Linux shards that run Rust tests, but not for our normal Linux shards.

I'm fine with adding it, although would err on the side of making as few changes as necessary to get green CI.

PR #35, opened 2019-03-14
Refactor ci.py to use ConfigParser

Eric-Arellano:
### Problem
While it's good https://github.com/pantsbuild/setup/pull/33 stepped up our CI game by introducing a new script and allowing us to now test an unspecified `pants_version`, we were reinventing the wheel with parsing and modifying `pants.ini`, which led to confusing code.

### Solution
Use Python's `ConfigParser` to greatly simplify and make more extendable `ci.py`.


cosmicexplorer (approved):
Thanks for splitting this out!!

PR #36, opened 2019-03-14
Refactor `./pants` and fix bugs in `ci.py` in preparation for Python 3 support

Eric-Arellano:
This makes several improvements and fixes that will result in a smaller diff for https://github.com/pantsbuild/setup/pull/32.

Fixes:
- Fix missing `write_config(updated_config)` when setting up `pants_version`, meaning when running `./ci.py --unspecified` we wouldn't actually remove the `pants_version` from `pants.ini`. This hotfixes https://github.com/pantsbuild/setup/pull/35.
- Properly restore `pants.ini` in CI, even if the subprocesses failed. Due to https://github.com/pantsbuild/setup/pull/34 now properly propagating return codes, the context manager would exit prematurely before restoring the original `pants.ini`.

Refactors:
- Check the Python interpreter is on the PATH and exit with error message if not.
- Extract Python interpreter setup into a dedicated function `determine_pants_runtime_python_version()` for a smaller diff with #32.
- Improve `pants.ini` parsing:
   - Extract `get_pants_ini_config_value()` to generify reading `pants.ini`.
   - Extend `pants.ini` parsing to recognize both `:` and `=`, along with optional spaces before and/or after the delimiter. These were all valid `ini` entries, but our script only handled one specific case.
   - Respect `#` in front of `pants_version` by requiring the line to start exactly with `pants_version`.
- Refactor `pants.ini` parsing to use variables for readability.
- Add `green` and `die` functions with colored output
   - Use `die` if interpreter cannot be resolved
   - Use `green` upon successful creation of a new virtual environment the first time

### DIFF
--- a/pants
+++ b/pants
@@ -48,18 +48,16 @@ function get_pants_ini_config_value {
   valid_delimiters="[:=]"
   optional_space="[[:space:]]*"
   prefix="^${config_key}${optional_space}${valid_delimiters}${optional_space}"
-  remove_prefix=" s#${prefix}##p"
-  sed -ne "/${prefix}/${remove_prefix}" pants.ini
+  sed -ne "/${prefix}/ s#${prefix}##p" pants.ini
 }
 
 function determine_pants_runtime_python_version {
   python_bin_name="${PYTHON:-python2.7}"
-  if which "${python_bin_name}" >/dev/null; then
-    echo "$(which ${python_bin_name})"
-  else
+  if ! which "${python_bin_name}" >/dev/null; then
     echo "${python_bin_name} could not be found on your PATH! Pants will not work." 1>&2
     exit 1
   fi
+  echo "$(which ${python_bin_name})"
 }
 
 # TODO(John Sirois): GC race loser tmp dirs leftover from bootstrap_XXX
@@ -113,7 +111,7 @@ function bootstrap_pants {
 # Ensure we operate from the context of the ./pants buildroot.
 cd "$(cd "$(dirname ${BASH_SOURCE[0]})" && pwd -P)"
 python="$(determine_pants_runtime_python_version)"
-pants_dir="$(bootstrap_pants ${python})"
+pants_dir="$(bootstrap_pants "${python}")"
 
 
 # We set the env var no_proxy to '*', to work around an issue with urllib using non
--- a/pants
+++ b/pants
@@ -57,7 +57,6 @@ function determine_pants_runtime_python_version {
     echo "Could not find ${python_bin_name}. Please ensure ${python_bin_name} is on your PATH." 1>&2
     exit 1
   fi
-  echo "$(which ${python_bin_name})"
 }
 
 # TODO(John Sirois): GC race loser tmp dirs leftover from bootstrap_XXX
--- a/pants
+++ b/pants
@@ -13,8 +13,6 @@
 
 set -eou pipefail
 
-PYTHON=${PYTHON:-$(which python2.7)}
-
 PANTS_HOME="${PANTS_HOME:-${XDG_CACHE_HOME:-$HOME/.cache}/pants/setup}"
 PANTS_BOOTSTRAP="${PANTS_HOME}/bootstrap-$(uname -s)-$(uname -m)"
 
@@ -45,6 +43,16 @@ function tempdir {
   mktemp -d "$1"/pants.XXXXXX
 }
 
+function determine_pants_runtime_python_version {
+	  python_bin_name="${PYTHON:-python2.7}"
+	  if which "${python_bin_name}" >/dev/null; then
+	    echo "$(which ${python_bin_name})"
+	  else
+	    echo "${python_bin_name} could not be found on your PATH! Pants will not work." 1>&2
+	    exit 1
+	  fi
+	}
+
 # TODO(John Sirois): GC race loser tmp dirs leftover from bootstrap_XXX
 # functions.  Any tmp dir w/o a symlink pointing to it can go.
 
@@ -64,6 +72,7 @@ function bootstrap_venv {
 }
 
 function bootstrap_pants {
+  python="$1"
   pants_requirement="pantsbuild.pants"
   pants_version=$(
     grep -E "^[[:space:]]*pants_version" pants.ini 2>/dev/null | \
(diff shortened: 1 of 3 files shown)


pants
```diff
@@ -45,12 +43,30 @@ function tempdir {
   mktemp -d "$1"/pants.XXXXXX
 }
 
+function get_pants_ini_config_value {
+  config_key="$1"
+  valid_delimiters="[:=]"
+  optional_space="[[:space:]]*"
+  prefix="^${config_key}${optional_space}${valid_delimiters}${optional_space}"
+  remove_prefix=" s#${prefix}##p"
+  sed -ne "/${prefix}/${remove_prefix}" pants.ini
```
cosmicexplorer: The way sed commands work in this case is that `/<pattern>/ s#<replace-pattern>#<replacement>#<flags>` says "for lines matching `<pattern>`, run the following `s` command -- replace `<replace-pattern>` in the line with `<replacement>`, printing or doing whatever according to `<flags>`". In this case I think the sed command itself shouldn't be split up, because the actual `s` command is what is doing things here and it makes it feel spooky otherwise. Something like:
```bash
  sed -ne "/${prefix}/ s#${prefix}##p" pants.ini
```
I think makes what's happening very clear ("when encountering prefix, remove it"). The use of `#` is something I like to do -- people tend to use `/` but that tends to clash a lot with things I want to replace, imo. The sed info pages are very nice but we hopefully shouldn't be doing anything complex enough to require them in a script we distribute like this.

pants
```diff
@@ -45,12 +43,30 @@ function tempdir {
...
+}
+
+function determine_pants_runtime_python_version {
+  python_bin_name="${PYTHON:-python2.7}"
+  if which "${python_bin_name}" >/dev/null; then
+    echo "$(which ${python_bin_name})"
+  else
+    echo "${python_bin_name} could not be found on your PATH! Pants will not work." 1>&2
+    exit 1
+  fi
```
cosmicexplorer, line 79: This is readable enough, but could also be:
```bash
  if ! which "${python_bin_name}"; then
    echo "${python_bin_name} could not be found on your PATH! Pants will not work." 1>&2
    exit 1
  fi
```
It will produce an extra `<x> not found` line in the error case, but I think that's ok / what we want here.

pants
```diff
@@ -65,40 +81,39 @@ function bootstrap_venv {
...
   fi
-  echo "${PANTS_BOOTSTRAP}/${pants_version}"
+  echo "${PANTS_BOOTSTRAP}/${target_folder_name}"
 }
 
 # Ensure we operate from the context of the ./pants buildroot.
 cd "$(cd "$(dirname ${BASH_SOURCE[0]})" && pwd -P)"
-pants_dir=$(bootstrap_pants)
+python="$(determine_pants_runtime_python_version)"
+pants_dir="$(bootstrap_pants ${python})"
```
```suggestion
pants_dir="$(bootstrap_pants "${python}")"
```
The inside of a `"$()"` substitution is a clean environment in terms of quoting, so the usual considerations with double-quoting dereferences apply. This isn't always necessary, but it could matter at some point, so it's useful to be consistent, I think.

Eric-Arellano: Didn't realize you can put multiple double quotes within double quotes! Good to know.

pants
Eric-Arellano: Yes I like this more. Thanks.

pants
Eric-Arellano: Agreed, this is much better. Thanks for the thorough explanation! Have not used sed much.

pants.ini
```diff
@@ -1,6 +1,6 @@
 [GLOBAL]
-pants_version:1.15.0.dev3
-plugins:[
+pants_version : 1.15.0.dev3
```
stuhood, line 2: Is this a related change? I've generally seen `key: value` , but presumably the matching code works for any combination.

pants
```diff
@@ -45,12 +43,28 @@ function tempdir {
   mktemp -d "$1"/pants.XXXXXX
 }
 
+function get_pants_ini_config_value {
+  config_key="$1"
+  valid_delimiters="[:=]"
+  optional_space="[[:space:]]*"
+  prefix="^${config_key}${optional_space}${valid_delimiters}${optional_space}"
```
stuhood, line 61: It looks like the old regex used to allow a leading optional space before the key... which is odd? It would seem to make more sense the way you have it here.

stuhood (approved):

pants.ini
Eric-Arellano: Yes, this unfortunately is the result of us using in `ci.py` `ConfigParser` to rewrite back our `pants.ini` after the tests. It doesn't allow us to specify the spacing beyond if we want no spaces or a space on both sides. 

Originally, our regex parsing wasn't flexible enough to allow a space before the `:` (even though it's perfectly valid `ini` syntax), so we had to use no spaces. Now that we can use either style, I changed back to the spaces because it's `ConfigParser`'s default and I think is easier to read.

pants
cosmicexplorer: With the above, I was actually thinking you could remove the `>/dev/null` and also remove the `echo` line below the `if`.

pants
```diff
@@ -65,40 +79,39 @@ function bootstrap_venv {
...
 function bootstrap_pants {
+  python="$1"
   pants_requirement="pantsbuild.pants"
-  pants_version=$(
-    grep -E "^[[:space:]]*pants_version" pants.ini 2>/dev/null | \
-      cut -f2 -d: | tr -d " "
-  )
-  if [[ -n "${pants_version}" ]]
-  then
+  pants_version=$(get_pants_ini_config_value 'pants_version')
```
```suggestion
  pants_version="$(get_pants_ini_config_value 'pants_version')"
```

cosmicexplorer (approved):
Thanks so much for iterating on this!

pants
```diff
@@ -33,24 +31,58 @@ SETUPTOOLS_REQUIREMENT="setuptools==5.4.1"
...
+# 3.) Check for a venv via a naming/path convention and execute if found.
+# 4.) Otherwise create venv and re-exec self.
 #
 # After that pants itself will handle making sure any requested plugins
 # are installed and up to date.
 
-function tempdir {
-  mktemp -d "$1"/pants.XXXXXX
+function determine_pants_runtime_python_version {
+  python_bin_name="${PYTHON:-python2.7}"
```
stuhood: Since `PYTHON` remains a "parameter" to this entire script, having it near the top of the file like it was before seems like good documentation of that fact.

stuhood (approved):

cosmicexplorer (approved):
Approving again for the `PYTHON_BIN_NAME` changes! Thanks!

PR #39, opened 2019-03-16
Cache pyenv in CI

Eric-Arellano:
### Problem
https://github.com/pantsbuild/setup/pull/33  resulted in us installing Python 3.6 on every shard but Trusty and Xenial, and https://github.com/pantsbuild/setup/pull/37 in us installing Python 3.7 on every OSX shard.

This slows down our CI unnecessarily and wastes time.

### Solution
Cache the pyenv folder for every shard, even if it does not use our custom Pyenv. This is safe to do because if the folder does not exist, Travis will simply upload an empty folder to the cache.

This requires adding a unique `CACHE_NAME` to every shard.

Further, we must refactor `.travis.yml` to globally define the env vars needed for our custom Pyenv install and to globally define `pyenv_install_py3{6,7}` for better readability.

## Result
Overall CI time reduced by 20.4 minutes and overall wall time by 6.1 minutes.


stuhood (dismissed):
Thanks.

.travis.yml
```diff
@@ -8,6 +8,9 @@ env:
   global:
     - PYENV_PY36_VERSION=3.6.8
     - PYENV_PY37_VERSION=3.7.2
+    - PYENV_ROOT="${PYENV_ROOT:-${HOME}/.pants_pyenv}"
```
jsirois, line 13: Why would you want to accept a pre-existing value here?

jsirois (approved):

.travis.yml
Eric-Arellano: Good question. Trusty and Xenial predefine `PYENV_ROOT` and overriding it means their Python 3 will no longer work. Added a comment to explain this, and will also update Pants repo to explain this as well.

PR #41, opened 2019-03-16
Add `build-support` and `common.py` to testing for better logging and easier CI development

Eric-Arellano:
### Problem
While `ci.py` has really nice benefits and makes it easy to parameterize our `pants.ini` environment, it was missing some nice utils our Pants repo's `ci.sh` uses, namely the logging functions like `start_travis_section` and `die`.

Further, `ci.py` had a couple functions that were util code, like `read_config()`, whereas the file should really only have logic related to tests and it should import those utils.

### Solution
* Move `ci.py` to `build-support/ci.py`, so that we have a place for test code and can easily in the future create things like `pre_commit.py`.
* Create a new `common.py` file, inspired by Pant's `common.sh`: https://github.com/pantsbuild/pants/blob/master/build-support/common.sh.
* Refactor `ci.py` to use these new utils.
   * Print how `pants_version` and `pants_runtime_python_version` are modified using `banner()`.
   * Use `with travis_section():` for each of the 4 tests
   * `die()` if `pants_version` missing from `pants.ini` when `--pants-version config` passed.

### Result
CI now has much better logging. See an example run: https://travis-ci.org/pantsbuild/setup/jobs/507333924.

We now have a better foundation for further improving CI, such as adding new scripts like `pre_commit.py`. This `common.py` will also serve as a foundation for rewriting some of our scripts in Pants to Python 3.


build-support/common.py
```diff
@@ -0,0 +1,98 @@
...
+from pathlib import Path
+from typing import Tuple
+
+# --------------------------------------------------------
+# Logging utils
+# --------------------------------------------------------
+
+SCRIPT_START_TIME = time.time()
+
+CLEAR_LINE = "\x1b[K"
```
stuhood: These constants should probably be private.

stuhood (approved):

PR #42, opened 2019-03-19
Refactor ci.py to allow passing multiple args in one call

Eric-Arellano:
### Problem
Typically, we want to run `ci.py` with multiple different environment configurations, such as ensuring it works with Python 2.7, 3.6, and 3.7. This currently means having to call the script again for each env.

This makes the script less flexible and results in more code in `.travis.yml`. It also makes the logs more verbose than necessary—ideally we would have a Travis fold for each new env, rather than the current behavior to fold for each individual subprocess run.

### Solution
Use argparse's `nargs` to allow this change.

Modify logging to have a Travis fold for each new env, rather than each individual subprocess run.

### Result
See https://travis-ci.org/pantsbuild/setup/jobs/508167593 for how logging looks now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 /.pids/
 
 __pycache__/
+/.mypy_cache/
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@
 /.pids/
 
 __pycache__/
-/.mypy_cache/
--- a/build-support/ci.py
+++ b/build-support/ci.py
@@ -3,6 +3,7 @@
 # Licensed under the Apache License, Version 2.0 (see LICENSE).
 
 import argparse
+import itertools
 import os
 import subprocess
 from contextlib import contextmanager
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -3,3 +3,4 @@
 /.pids/
 
 __pycache__/
+/.mypy_cache/
```
cosmicexplorer: Was this necessary before as well?

cosmicexplorer (approved):
Moving code out of `.travis.yml` is a win! This seems much easier to read as well.

.gitignore
Eric-Arellano: Sort of. The repo never runs MyPy, but I've been running it locally through system installed MyPy and kept accidentally adding the cache to Git.

Do you prefer I remove this for a cleaner diff?

cosmicexplorer: If it's not going to be generated for all users of the repo / we never run the mypy command in any scripts in this repo then I might suggest considering [`.git/info/exclude`](https://stackoverflow.com/questions/22906851/when-would-you-use-git-info-exclude-instead-of-gitignore-to-exclude-files) -- I use it for e.g. `.dir-locals.el` files which emacs uses.

build-support/ci.py
```diff
@@ -32,46 +33,62 @@ def __str__(self):
...
 
 
+def run_tests_with_env(*,
+    pants_version: PantsVersion, python_version: PythonVersion, skip_pantsd_tests: bool
+  ) -> None:
+  slug = f"Tests_{pants_version}_{python_version}"
+  banner_message = f"Running tests with `--pants-version={pants_version}` and `--python_version={python_version}`."
+  with travis_section(slug, banner_message):
+    with setup_pants_version(pants_version), setup_python_version(python_version):
+      run_tests(skip_pantsd_tests=skip_pantsd_tests)
```
cosmicexplorer, line 76: In the pants repo, I typically see this done as:
```python
 with travis_section(slug, banner_message),\
      setup_pants_version(pants_version),\
      setup_python_version(python_version):
    run_tests(skip_pantsd_tests=skip_pantsd_tests)
```
and the vertical alignment feels easier for me to read, personally. But using multiple `with`s is definitely also useful to demarcate logically separate contexts.

PR #43, opened 2019-03-19
Add shellcheck.py script and improve ./pants with results

Eric-Arellano:
### Problem
The entire point of this repo is the `./pants` script, so we should be using as much CI as possible to ensure it works well. While we do now have `ci.py` to act as an integration test, we would benefit from a linter.

### Solution
Use `shellcheck`, which is the most used and recommended linter for Bash scripts. 

Specifically:
* Add new `shellcheck.py` script
* Add new CI shard that runs `shellcheck.py` to avoid regressions
* Fix issues discovered in `./pants`

### DIFF
--- a/build-support/shellcheck.py
+++ b/build-support/shellcheck.py
@@ -15,8 +15,8 @@ def main() -> None:
 
 def ensure_shellcheck_installed() -> None:
   if shutil.which("shellcheck") is None:
-    die("`shellcheck` not installed! You may download this through brew or apt. "
-        "See https://www.shellcheck.net.")
+    die("`shellcheck` not installed! You may download this through brew, apt, or yum. "
+        "See https://github.com/koalaman/shellcheck#installing.")
 
 
 def run_shellcheck() -> None:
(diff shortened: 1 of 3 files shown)


pants
```diff
@@ -147,6 +147,6 @@ pants_dir="$(bootstrap_pants "${python}")"
 # async-signal-safe syscalls after we fork a process that has already spawned threads.
 #
 # See https://blog.phusion.nl/2017/10/13/why-ruby-app-servers-break-on-macos-high-sierra-and-what-can-be-done-about-it/
-export no_proxy=*
+export no_proxy='*'
```
cosmicexplorer, line 150: Very useful change!

build-support/shellcheck.py
```diff
@@ -0,0 +1,33 @@
...
+
+def main() -> None:
+  ensure_shellcheck_installed()
+  run_shellcheck()
+
+
+def ensure_shellcheck_installed() -> None:
+  if shutil.which("shellcheck") is None:
+    die("`shellcheck` not installed! You may download this through brew, apt, or yum. "
+        "See https://github.com/koalaman/shellcheck#installing.")
```
cosmicexplorer: I would recommend:
> `shellcheck` not installed! You may download this through your operating system's package manager, such as brew, apt, or yum. See https://github.com/koalaman/shellcheck#installing.

I use pacman on arch linux, for example.

cosmicexplorer (approved):

PR #45, opened 2019-03-21
Auto-generate pants.ini with pinned pants_version if file is missing

Eric-Arellano:
### Problem
Our first-time [setup instructions](https://www.pantsbuild.org/install.html#recommended-installation) are harder than need be. Currently, before being able to do anything with Pants, you must:

1) curl `pants`
1) `chmod +x pants`
1) `touch pants.ini`
1) Copy and paste `[GLOBAL]` into the file.
1) Run `./pants -V` or decide you want to pin a different value.
1) Copy and paste that value into a new line `pants_version`.

Instead, we could simplify the instructions to:

1) curl `pants`
1) `chmod +x pants`
1) `./pants`
1) Inspect `pants.ini` to make sure you're okay with the defaults.

Simplifying the onboarding process should make it easier for us to retain first time users. There are a lot of new concepts they already have to learn, e.g. BUILD files, targets, and goals. The less overwhelming we can make this process, the better.

Refer to https://pantsbuild.slack.com/archives/CBNMV1LRH/p1552506655047900 for the original discussion around this idea.

#### Concern about creating `pants.ini` being a tutorial
While there is some value in creating `pants.ini` so that users see how the config works, we can still get this benefit via logging when we create it and then instructing them on the website to inspect the file and change the values if desired.

### Solution
* Add logic to check if `pants.ini` exists and create it if not.
* Modify `pants_ini_config_value` to work if `pants.ini` is missing.

#### Testing
Add a test file specifically dedicated to this feature that temporarily deletes our `pants.ini`, runs `./pants`, and inspects the results to confirm they're as expected.

This is a separate file, rather than being with `ci.py`, because they are testing different things. This solely is meant to test our auto-generation works, whereas `ci.py` is meant to test that we 1) bootstrap properly, 2) properly parse `pants_version` and `pants_runtime_python_version`, and 3) `./pants -V` and `./pants list ::` work with both pantsd on and off.

### Result
If you already have `pants.ini`, everything behaves the same.

If you don't, we'll create it and print this:

```
Config file `pants.ini` not detected in the repository root. Creating with sensible defaults:

* Pinning `pants_version` to the most recent stable release: 1.14.0.

`pants.ini` created. You may now run `./pants` normally.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,7 +91,7 @@ matrix:
           --skip-pantsd-tests
         - ./build-support/ci.py
           --pants-versions config
-          --python-versions unspecified 2.7 3.6
+          --python-versions unspecified 2.7 3.6 3.7
           --skip-pantsd-tests
         - ./build-support/test_pants_ini_autogen.py
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ script:
   - ./build-support/ci.py
     --pants-versions config
     --python-versions unspecified 2.7 3.6 3.7
+  - ./build-support/test_pants_ini_autogen.py
 
 osx_setup: &osx_setup
   os: osx
@@ -92,6 +93,7 @@ matrix:
           --pants-versions config
           --python-versions unspecified 2.7 3.6
           --skip-pantsd-tests
+        - ./build-support/test_pants_ini_autogen.py
 
 
     - name: "OSX 10.13 - High Sierra"
@@ -114,6 +116,7 @@ matrix:
         - ./build-support/ci.py
           --pants-versions config
           --python-versions unspecified 2.7 3.6
+        - ./build-support/test_pants_ini_autogen.py
 
     - name: "Ubuntu 14.04 - Trusty"
       <<: *linux_setup
@@ -127,6 +130,7 @@ matrix:
         - ./build-support/ci.py
           --pants-versions config
           --python-versions unspecified 2.7 3.6
+        - ./build-support/test_pants_ini_autogen.py
 
 
     - name: "Ubuntu 16.04 - Xenial"
--- a/build-support/common.py
+++ b/build-support/common.py
@@ -100,9 +100,9 @@ def temporarily_rewrite_config(updated_config: configparser.ConfigParser):
 
 @contextmanager
 def temporarily_remove_config():
-  original_config = read_config()
-  Path(PANTS_INI).unlink()
+  Path(PANTS_INI).rename("pants.ini.orig")
   try:
     yield
   finally:
-    write_config(original_config)
+    Path(PANTS_INI).unlink()
+    Path("pants.ini.orig").rename("pants.ini")
--- a/build-support/common.py
+++ b/build-support/common.py
@@ -96,3 +96,13 @@ def temporarily_rewrite_config(updated_config: configparser.ConfigParser):
     yield
   finally:
     write_config(original_config)
+
+
+@contextmanager
+def temporarily_remove_config():
+  original_config = read_config()
+  Path(PANTS_INI).unlink()
+  try:
+    yield
+  finally:
+    write_config(original_config)
--- a/build-support/test_pants_ini_autogen.py
+++ b/build-support/test_pants_ini_autogen.py
@@ -1,0 +1,42 @@
+#!/usr/bin/env python3
+# Copyright 2019 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+import configparser
+import subprocess
+from pathlib import Path
+
+from ci import PantsVersion, setup_pants_version
+from common import (CONFIG_GLOBAL_SECTION, PANTS_INI, die, read_config,
+                    temporarily_remove_config, travis_section)
+
+
+def main() -> None:
+  with travis_section("PantsIniAutoGen", "Testing auto-generation of pants.ini."):
+    with temporarily_remove_config():
+      subprocess.run(["./pants"], check=True)
+      assert_file_created()
+      config = read_config()
+      assert_pants_version_pinned_properly(config)
+
+
+def assert_file_created() -> None:
+  if not Path(PANTS_INI).is_file():
+    die("pants.ini not created!")
+
+
+def assert_pants_version_pinned_properly(config: configparser.ConfigParser) -> None:
+  if "pants_version" not in config[CONFIG_GLOBAL_SECTION]:
+    die("`pants_version` not pinned!")
+  pinned_pants_v = config[CONFIG_GLOBAL_SECTION]["pants_version"]
+  with setup_pants_version(PantsVersion.unspecified):
+    unconfigured_pants_v = subprocess.run(
+      ["./pants", "-V"], stdout=subprocess.PIPE, encoding="utf-8", check=True
+    ).stdout.strip()
+  if pinned_pants_v != unconfigured_pants_v:
+    die(f"The pinned `pants_version` ({pinned_pants_v}) does not match the value when "
+        f"leaving `pants_version` unspecified ({unconfigured_pants_v})")
+
+
+if __name__ == "__main__":
+  main()
--- a/build-support/test_pants_ini_autogen.py
+++ b/build-support/test_pants_ini_autogen.py
@@ -22,12 +22,12 @@ def main() -> None:
 
 def assert_file_created() -> None:
   if not Path(PANTS_INI).is_file():
-    die("pants.ini not created!")
+    die("pants.ini not created in the repo root.")
 
 
 def assert_pants_version_pinned_properly(config: configparser.ConfigParser) -> None:
   if "pants_version" not in config[CONFIG_GLOBAL_SECTION]:
-    die("`pants_version` not pinned!")
+    die("`pants_version` not pinned.")
   pinned_pants_v = config[CONFIG_GLOBAL_SECTION]["pants_version"]
   with setup_pants_version(PantsVersion.unspecified):
     unconfigured_pants_v = subprocess.run(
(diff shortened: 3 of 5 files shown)


.travis.yml
```diff
@@ -90,8 +91,9 @@ matrix:
           --skip-pantsd-tests
         - ./build-support/ci.py
           --pants-versions config
-          --python-versions unspecified 2.7 3.6
+          --python-versions unspecified 2.7 3.6 3.7
```
Eric-Arellano, line 94: This was accidentally removed in https://github.com/pantsbuild/setup/pull/42.

pants
```diff
@@ -137,10 +141,27 @@ function bootstrap_pants {
...
 }
 
+function create_pants_ini_if_missing {
+  if [[ -f "pants.ini" ]]; then
+    return 0
+  fi
+  log "Config file \`pants.ini\` not detected in the repository root. Creating with sensible defaults.\n"
+  touch "pants.ini"
+  pants_version="$(./pants -V)"
+  log "* Pinning \`pants_version\` to the most recent stable release: ${pants_version}."
```
Eric-Arellano, line 148: After 1.15.0 is released, we'll also automatically pin `pants_runtime_python_version`!

```
* Pinning `pants_runtime_python_version` to Python 2.7. Ensure you have this interpreter available everywhere you plan to use Pants!
```

Will probably try to default to Python 3.6, then 3.7, then 2.7, so that dropping Py27 after 1.17.0 is a bit less surprising. TBD..

build-support/common.py
```diff
@@ -96,3 +96,13 @@ def temporarily_rewrite_config(updated_config: configparser.ConfigParser):
...
+
+
+@contextmanager
+def temporarily_remove_config():
+  original_config = read_config()
+  Path(PANTS_INI).unlink()
+  try:
+    yield
+  finally:
+    write_config(original_config)
```
cosmicexplorer: I would temporarily rename the file to something like `.pants.ini.orig` instead of deleting it here, mostly on principle.

build-support/test_pants_ini_autogen.py
```diff
@@ -0,0 +1,48 @@
...
+                    temporarily_remove_config, travis_section)
+
+
+def main() -> None:
+  with travis_section("PantsIniAutoGen", "Testing auto-generation of pants.ini."):
+    with temporarily_remove_config():
+      subprocess.run(["./pants"], check=True)
+      assert_file_created()
+      config = read_config()
+      assert_pants_version_pinned_properly(config)
```
cosmicexplorer: Is there a reason this file couldn't be a `unittest.TestCase`?

pants
cosmicexplorer: Would it be even less surprising if we were to also mention in the logging to the user here that in 1.15.0 we will automatically pin `pants_runtime_python_version`?

build-support/test_pants_ini_autogen.py
```diff
@@ -0,0 +1,48 @@
...
+
+def assert_file_created() -> None:
+  if not Path(PANTS_INI).is_file():
+    die("pants.ini not created in the repo root.")
+
+
+def assert_pants_version_pinned_properly(config: configparser.ConfigParser) -> None:
+  if "pants_version" not in config[CONFIG_GLOBAL_SECTION]:
+    die("`pants_version` not pinned.")
+  pinned_pants_v = config[CONFIG_GLOBAL_SECTION]["pants_version"]
```
cosmicexplorer: Would recommend avoiding abbreviating variable names here since the meaning isn't entirely unambiguous imho.

pants
```diff
@@ -137,10 +141,27 @@ function bootstrap_pants {
   echo "${PANTS_BOOTSTRAP}/${target_folder_name}"
 }
 
+function create_pants_ini_if_missing {
+  if [[ -f "pants.ini" ]]; then
+    return 0
+  fi
+  log "Config file \`pants.ini\` not detected in the repository root. Creating with sensible defaults.\n"
+  touch "pants.ini"
```
cosmicexplorer, line 146: I think you can remove this `touch`, just add more `log "* ..."` lines for any other sensible defaults we may add in the future, and still write it all at once in a heredoc.

pants
```diff
@@ -137,10 +141,27 @@ function bootstrap_pants {
   echo "${PANTS_BOOTSTRAP}/${target_folder_name}"
 }
 
+function create_pants_ini_if_missing {
+  if [[ -f "pants.ini" ]]; then
+    return 0
+  fi
+  log "Config file \`pants.ini\` not detected in the repository root. Creating with sensible defaults.\n"
```
```suggestion
  log 'Config file `pants.ini` not detected in the repository root. Creating with sensible defaults:'
```
As mentioned below, I personally try to use single quotes where possible to make it clear where things aren't getting substituted, but it's not a necessary change. I would personally use a colon instead of a period, though, and I feel like it makes more sense to then drop the newline.

pants
```diff
@@ -137,10 +141,27 @@ function bootstrap_pants {
...
+  touch "pants.ini"
+  pants_version="$(./pants -V)"
+  log "* Pinning \`pants_version\` to the most recent stable release: ${pants_version}."
+  cat <<EOF >> "pants.ini"
+[GLOBAL]
+pants_version: ${pants_version}
+EOF
+  green "\n\`pants.ini\` created. You may now run \`./pants\`."
```
cosmicexplorer, line 153: Within single quotes, nothing is substituted (which is why I try to use them where possible, to make it more clear when things *are* getting substituted into a string), so you can do things a little more fearlessly:
```suggestion
  green '\n`pants.ini` created. You may now run `./pants`.'
```

pants
```diff
@@ -137,10 +141,27 @@ function bootstrap_pants {
...
+  exit 0
```
cosmicexplorer: Calling `exit` within a function unless it's something like `die` can be a bit hard to follow. I think below (instead of `create_pants_ini_if_missing`) could be something like:

```bash
if [[ ! -f 'pants.ini' ]]; then
  autogen_pants_ini
  exit 0
fi
```
and then we can remove the `exit` here and the check at the top of this method too.

cosmicexplorer (commented):
I am a huge fan of this change! I personally always felt annoyed that I had to go look up how to set up pants every single time I wanted to make a new repo and I think being explicit about the changes we're doing automatically (but still doing them automatically) is a great precedent to set.

pants
Eric-Arellano: We have to keep the touch because of the recursive call to `./pants -V`. Otherwise the script will attempt to continually generate the file.

pants
Eric-Arellano: I think that might confuse people here, because there is nothing actionable for them to be able to do.

We're only about 1-3 weeks away from 1.15.0, depending on how many RCs we have, so this will be fixed soon!

pants
Eric-Arellano: I tried this suggestion, but then `shellcheck.py` complains that it thinks we're intending to do command substitution and it explains we would then need double quotes. So I kept as is.

build-support/common.py
Eric-Arellano: Done!

build-support/test_pants_ini_autogen.py
Eric-Arellano: No good reason at all, beyond not knowing how to run unit tests as a script. It was important to me that we don't make the caller do something like `python3 -m unittest ...`.

Figured this out though!
```python
if __name__ == "__main__":
  with travis_section("PantsIniAutoGen", "Testing auto-generation of pants.ini."):
    unittest.main()
```

jsirois:
Did you think about instead making this a goal so the code / complication could live in the pants repo?. The instructions might just be:
1. curl
2. chmod
3. `./pants version`

Where `version` might print out the version like `--version` when configured and otherwise prompt to pick the version and edit pants.ini if unset.

Eric-Arellano:
> Did you think about instead making this a goal so the code / complication could live in the pants repo?. The instructions might just be:

No, I did not think much about making it a Pants goal. 

My main concern would be that after 1.15.0, we will also wan to auto-pin `pants_runtime_python_version`. That wouldn't make as much sense to live in the Pants repo, as Pants cannot / should not modify the Python interpreter it runs with, since that is the responsibility of the calling `./pants` script.

Also, this feature is solely for first time users who do not already have `pants.ini`, so it seems less right for this to belong in Pants repo than in a setup script.

stuhood:
I agree with @jsirois that it is worth considering at least. If you still wanted it to be automagic, you could replace his third step with something like `./pants pin-pants-version`, which would write out the file or prompt you if it already exists.

Eric-Arellano:
> I agree with @jsirois that it is worth considering at least. If you still wanted it to be automagic, you could replace his third step with something like `./pants pin-pants-version`, which would write out the file or prompt you if it already exists.

This would still require the user to still create `pants.ini`, or `./pants pin-pants-version` will fail to work.

Steps would be:
1. curl `pants`
2. `chmod +x pants`
3. `touch pants.ini`
4. Copy and paste `[GLOBAL]` into the file.
5. `./pants pin-pants-version`.

Maybe we could skip step 4 and have `./pants pin-pants-version` add `[GLOBAL]` if missing, in which case it would be:
1. curl `pants`
2. `chmod +x pants`
3. `touch pants.ini`
4. `./pants pin-pants-version`.

--

Once we add pinning `pants_runtime_python_version`, steps would be:
1. curl `pants`
2. `chmod +x pants`
3. `touch pants.ini`
4. `./pants pin-pants-version`.
5. `./pants pin-pants-runtime-python-version`.

As opposed to:
1. curl `pants`
2. `chmod +x pants`
3. `./pants`, which would log everything being pinned.

--

If we go with the goals approach, would we allow optional arguments?

```bash
$ ./pants pin-pants-version 1.15.0.dev4
$ ./pants pin-pants-runtime-python-version 3.6
```

If left off, default to current version being used and current interpreter being used?

--

Any risk for slippery slope, e.g. why do we not have `pin-python-setup-interpreter-constraints` or `pin-x` etc?

--

Both options are improvements on what we have, so I can take the Pants goals approach if you all think it's the right approach. Let me know.

pants
cosmicexplorer: Ah, great!

Eric-Arellano:
Okay I talked with Danny and agree with you both now! We came up with this proposal:

```bash
./pants generate-config
```

Defaults to using the Pants Version in the file `VERSION` and the Python interpreter being used.

--

Also would allow optional args:

```
./pants generate-config --new-pants-version 1.15.0.dev4 --new-runtime-python-version 3.6
```

If pants.ini already exists, it will modify it with those new values and then exit, so they get picked up the next Pants invocation.

--

This has the benefit of being a tutorial to running Pants and how goals work.

Also addresses my main concern of a proliferation of `pin-x` goals.

PR #46, opened 2019-03-21
Redirect to pantsbuild.org in README and index.html

Eric-Arellano:
We were duplicating the instructions. It would be better to point to the official website in the README and to automatically redirect with the site.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,11 +1,5 @@
 <html>
   <head>
-    <title>Pants Setup</title>
+    <meta http-equiv="refresh" content="0; url=https://www.pantsbuild.org/install">
   </head>
-  <body>
-    <p style="font-size:150%;">
-    Follow the instructions at <a href="https://www.pantsbuild.org/install">
-      https://www.pantsbuild.org/install</a> to setup Pants.
-    </p>
-  </body>
 </html>
--- a/index.html
+++ b/index.html
@@ -1,5 +1,12 @@
 <html>
   <head>
+    <title>Pants Setup</title>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -3,46 +3,9 @@
...
-    Pants will notice you changed your plugins and install them.
-    </p>
-    <p>
-    NB: The formatting of the plugins list is important; all lines below the
-    `plugins:` line must be indented by at least one white space to form logical
-    continuation lines.  This is standard for python ini files, see
-    <a href="http://tools.ietf.org/html/rfc822.html#section-3.1">RFC #822</a>
-    section 3.1.1 for the full rules python uses to parse ini file entries.
+    <p style="font-size:150%;">
+    Follow the instructions at <a href="https://www.pantsbuild.org/install">
```
jsirois, line 8: This is fine, but adding an actual redirect too would be best. I think the following is robustly cross-browser compatible:
```html
<html>
  <head>
    <meta http-equiv="refresh" content="0; url=https://www.pantsbuild.org/install">
  </head>
  ...
</html>
```

jsirois (approved):

index.html
Eric-Arellano: Ah much better! Thank you.

index.html
```diff
@@ -1,11 +1,5 @@
 <html>
   <head>
-    <title>Pants Setup</title>
```
jsirois: It probably would be best to keep the LHS and just add the meta tag. If - for whatever reason - a refresh does not occur - having a title and and explanation of where to go is a good thing.

jsirois (approved):

PR #48, opened 2019-03-29
Fix ci.py messing up the formatting of pants.ini

Eric-Arellano:
### Problem
ConfigParser formats things strangely when writing to a file. So, any call to `ci.py` would result in our `pants.ini` permanently having bad formatting that conflicts with how we things at https://www.pantsbuild.org/install.html.

### Solution
With the context manager `temporarily_rewrite_config`, simply grab the original content and rewrite it back exactly, rather than using ConfigParser to write back.

Also add back testing of Python 3.7 to the OSX 10.12 shard, as it was unintentionally removed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,7 +90,7 @@ matrix:
           --skip-pantsd-tests
         - ./build-support/ci.py
           --pants-versions config
-          --python-versions unspecified 2.7 3.6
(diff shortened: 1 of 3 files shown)


cosmicexplorer (approved):
Thanks for fixing this! I remember seeing this earlier but not wanting to slow down the great work happening in this repo -- thanks for following up with a fix!

PR #49, opened 2019-03-29
Determine default Python interpreter based on the pants_version being used

Eric-Arellano:
### Problem
Now that using Python 2.7 is ~deprecated and support will be removed soon, we need to stop defaulting to 2.7. 

Defaulting to Python 3 will make for a less stark transition. Specifically, users will be able to leave out `pants_runtime_python_version` from their `pants.ini`, which is good because we will deprecate this option once we drop Py2.

However, we do not want to break backwards compatibility. A user should be able to downgrade to 1.14.0 without issues.

### Solution
Determine which `pants_version` the user has pinned, and set the default interpreter from there, as follows:

* `<= 1.14.0`: Python 2.7
* `== 1.15.0`: Python 3.6->2.7
* `== 1.16.0`: Python 3.6->3.7->2.7
* `== 1.17.0`: Python 3.6->3.7->3.8->3.9
* `> 1.17.0`: Python 3.9->3.8->3.7->3.6

Notably, we first default to Python 3 and only fall back to Python 2 when necessary. We do this because we want most people using Python 3 so that we can completely stress test it before dropping Python 2.

For versions `> 1.17.0`, we default to using the max interpreter possible and only fall back to 3.6 as a last resort. This allows us to make no assumptions about Pants supporting 3.6 in the future.

If the user has not pinned `pants_version`, such as happens with first time users, we use the default interpreters for the current major release. With 1.15.0, this means we would use 3.6 or 2.7. Until 1.18.0 is released and our migration is stabilized, we will update the default case in the `./pants` script to match that new release. This should only impact new users before they run the `./pants generate-pants-ini` command.

We also avoid the ambiguous `python3` bin name to avoid accidentally trying to use Python 3.5, for example.

### DIFF
--- a/pants
+++ b/pants
@@ -90,7 +90,7 @@ function determine_default_python_exe {
     supported_python_versions=('3.6' '2.7')
     supported_message='2.7 or 3.6'
   else
-    pants_minor_version="$(echo ${pants_version} | cut -d '.' -f2)"
+    pants_minor_version="$(echo "${pants_version}" | cut -d '.' -f2)"
     if [[ "${pants_minor_version}" -ge 17 ]]; then
       supported_python_versions=('3.6' '3.7' '3.8' '3')
       supported_message='3.6+'
@@ -108,7 +108,7 @@ function determine_default_python_exe {
   for version in "${supported_python_versions[@]}"; do
     command -v "python${version}" && return 0
   done
-  die "No valid valid Python interpreter found. Pants requires Python ${supported_message}."
+  die "No valid Python interpreter found. Pants requires Python ${supported_message}."
 }
 
 function determine_python_exe {
--- a/pants
+++ b/pants
@@ -93,8 +93,8 @@ function determine_default_python_exe {
     pants_major_version="$(echo "${pants_version}" | cut -d '.' -f1)"
     pants_minor_version="$(echo "${pants_version}" | cut -d '.' -f2)"
     if [[ "${pants_major_version}" -eq 1 && "${pants_minor_version}" -le 14 ]]; then
-      supported_python_versions=('3.6' '2.7')
-      supported_message='2.7 or 3.6'
+      supported_python_versions=('2.7')
+      supported_message='2.7'
     elif [[ "${pants_major_version}" -eq 1 && "${pants_minor_version}" -eq 15 ]]; then
       supported_python_versions=('3.6' '2.7')
       supported_message='2.7 or 3.6'
--- a/pants
+++ b/pants
@@ -92,27 +92,21 @@ function determine_default_python_exe {
   else
     pants_major_version="$(echo "${pants_version}" | cut -d '.' -f1)"
     pants_minor_version="$(echo "${pants_version}" | cut -d '.' -f2)"
-    if [[ "${pants_major_version}" -eq 2 ]]; then
-      supported_python_versions=('3.9' '3.8' '3.7' '3.6')
+    if [[ "${pants_major_version}" -eq 1 && "${pants_minor_version}" -le 14 ]]; then
+      supported_python_versions=('3.6' '2.7')
+      supported_message='2.7 or 3.6'
+    elif [[ "${pants_major_version}" -eq 1 && "${pants_minor_version}" -eq 15 ]]; then
+      supported_python_versions=('3.6' '2.7')
+      supported_message='2.7 or 3.6'
+    elif [[ "${pants_major_version}" -eq 1 && "${pants_minor_version}" -eq 16 ]]; then
+      supported_python_versions=('3.6' '3.7' '2.7')
+      supported_message='2.7, 3.6, or 3.7'
+    elif [[ "${pants_major_version}" -eq 1 && "${pants_minor_version}" -eq 17 ]]; then
+      supported_python_versions=('3.6' '3.7' '3.8' '3.9')
       supported_message='3.6+'
     else
-      if [[ "${pants_minor_version}" -gt 17 ]]; then
-        supported_python_versions=('3.9' '3.8' '3.7' '3.6')
-        supported_message='3.6+'
-      elif [[ "${pants_minor_version}" -eq 17 ]]; then
-        supported_python_versions=('3.6' '3.7' '3.8' '3.9')
-        supported_message='3.6+'
-      elif [[ "${pants_minor_version}" -eq 16 ]]; then
-        supported_python_versions=('3.6' '3.7' '2.7')
-        supported_message='2.7, 3.6, or 3.7'
-      elif [[ "${pants_minor_version}" -eq 15 ]]; then
-        supported_python_versions=('3.6' '2.7')
-        supported_message='2.7 or 3.6'
-      # <= 1.14.0 does not support Python 3.
-      else
-        supported_python_versions=('2.7')
-        supported_message='2.7'
-      fi
+      supported_python_versions=('3.9' '3.8' '3.7' '3.6')
+      supported_message='3.6+'
     fi
   fi
   for version in "${supported_python_versions[@]}"; do
--- a/pants
+++ b/pants
@@ -73,7 +73,8 @@ EOF
 
 # The high-level flow:
 # 1.) Resolve the Python interpreter, first reading from the env var $PYTHON,
-#     then reading from pants.ini, then defaulting to python3.6, and finally to python2.7.
+#     then reading from pants.ini, then determining the default based off of
+#     which Python versions the current `pants_version` supports.
 # 2.) Grab pants version from pants.ini or default to latest.
 # 3.) Check for a venv via a naming/path convention and execute if found.
 # 4.) Otherwise create venv and re-exec self.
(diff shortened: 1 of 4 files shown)


jsirois:
I've held off review since the CI failures look legit and I can't suss the bug reading the diff.

Eric-Arellano:
Ah my bad - should have mentioned I expect CI to fail until 1.15.0 is released. Pants doesn't work with Python 3 when using 1.14.0, which is what an unconfigured `pants_version` defaults to.

Note everything passes except for `--pants-version unspecified --python-version unspecified`.

--

My intent in requesting a review was so that this is ready for me to merge once 1.15.0 is released and we get green CI.

jsirois (approved):

Eric-Arellano:
Important clarification with this PR. Users who already have the script won't be impacted by the change, unless they curl the script again. This is a good thing, so that we don't change their behavior for 1.15.0 overnight.

Now, thanks to this PR, we will be able to simplify our instructions for current users to upgrade to Python 3 to simply:
1. `curl -L -O https://pantsbuild.github.io/setup/pants` for the changes we made to `./pants` to default to using Python 3.6 and then to fall back to Python 2.7.
2. Change the `pants_version` in pants.ini to 1.15.0.
3. Run `./pants`.
4. Save your changes to `./pants` and `pants.ini` to your organization's version control.

stuhood (commented):
Is it possible (and/or worthwhile) to choose the default based on which side of the `1.15.0` split you are on? Presumably that would require the script to parse pants version numbers though.

Eric-Arellano:
@stuhood good idea! I would propose not only checking against 1.15.0, but also 1.16.0 and 1.17.0 splits, as follows:

```
if pants_version <= 1.14.0.x:
  python 2.7 -> die
elif pants_version == 1.15.0.x:
  python 3.6 -> python2.7 -> die
elif pants_version == 1.16.0.x:
  python 3.6 -> python 3.7 -> python2.7 -> die
elif pants_version >= 1.17.0.x:
  python3.6 -> python3.7 -> python3.8 -> python3 -> die
else:
  python 3.6 -> python2.7 -> die
```

At the cost of more complex code, we would have significantly less churn in this repo. Because people rarely update their `./pants` script, I think that is a major benefit. 

For example, with this scheme, new users who download the script at 1.16.0 will no longer have the risk of the `./pants` script trying to use Python 2.7 when they upgrade to 1.17.0.

-- 

Foil here though, what to do when `pants_version` is not defined? When undefined, Pip determines which `pants_version` to use and we definitely don't want to rely on that result to dynamically change the default interpreter.

With this question, note that we no longer pin `pants_runtime_python_version` when running `./pants generate-pants-ini`, so the interpreter we use that first time is a bit less relevant.

--

@jsirois is this getting too complex, do you think? Or is the reduction in churn + allowing users to downgrade versions without issues worth it?

stuhood:
> When undefined, Pip determines which pants_version to use and we definitely don't want to rely on that result to dynamically change the default interpreter.

Hm... I suspect that we do? Pip defaults to "the latest stable version", which is a good default, and a reasonable way to choose the default python as well.

Eric-Arellano:
> When undefined, Pip determines which pants_version to use and we definitely don't want to rely on that result to dynamically change the default interpreter.
> 
> Hm... I suspect that we do? Pip defaults to "the latest stable version", which is a good default, and a reasonable way to choose the default python as well.

The issue is technical with the execution order.

https://github.com/pantsbuild/setup/blob/336348199da772cf4995e27aca238479f3868c09/pants#L74-L79

Pip requires the virtual env to already be set up, and pip corresponds to a specific Python version. Let's say 1.17.0 is now the stable release—then trying to use pip2 to install Pants and subsequently to determine which `pants_version` we're on will fail.

--

Instead, what I propose is based on the assumption that the majority of users will have `pants_version` pinned. In both our old and new documentation, the very first thing we say to do is to pin `pants_version`.

Given this assumption, the only time this undefined `pants_version` case matters is for first time users before they run `./pants generate-pants-ini`. After that first run, they will have `pants_version` defined so above behavior means Pants will always work regardless of their `pants_version`.

So, for the undefined `pants_version` case, I propose that we change the defaults as we release each stable major version. That means in 1.15.0, we change it to `python3.6->python2.7->die`, then in 1.16.0 we change the script again to `python3.6->python3.7->python2.7->die`, and so on. 

While this has some churn in the script, the key detail is pre-existing users have no need to re-curl the script. These updates are only expected to impact first-time users.

pants
```diff
@@ -73,23 +73,54 @@ EOF
...
+      supported_message='2.7 or 3.6'
+    else
+      supported_python_versions=('2.7')
+      supported_message='2.7'
+    fi
+  fi
+  for version in "${supported_python_versions[@]}"; do
+    command -v "python${version}" && return 0
+  done
+  die "No valid valid Python interpreter found. Pants requires Python ${supported_message}."
```
stuhood: valid valid

Also, is it worth including the pants version in this error message?

pants
```diff
@@ -73,23 +73,54 @@ EOF
...
-function determine_pants_runtime_python_version {
+function determine_default_python_exe {
+  pants_version="$(get_pants_ini_config_value 'pants_version')"
+  # If `pants_version` not pinned in `pants.ini`, default to what the most
+  # recent stable release supports.
+  if [[ -z "${pants_version}" ]]; then
+    supported_python_versions=('3.6' '2.7')
+    supported_message='2.7 or 3.6'
+  else
+    pants_minor_version="$(echo ${pants_version} | cut -d '.' -f2)"
```
stuhood: If the major version is greater than `1`, we should default to the max python versions. Just in case we ever cut `2.0.0` =)

stuhood (commented):
Thanks!

pants
Eric-Arellano: This would require adding even more onto this already long function so that we parse the major version, in addition to the minor version.

I'd advocate for us to leave this out. We don't know yet when 2.0 will be cut nor which Python versions will be out when that happens. Also, I think it's totally reasonable for us to request once 2.0 lands that users curl the script again, just as we're making them do for Py3 upgrade.

Still, 2.0 will be exciting :) maybe once we use V2 completely?

pants
Eric-Arellano: Good catch.

I'm not sure about including `pants_version`. It's useful when it's pinned, although we do not know the `pants_version` when it is not pinned. We could keep it generic and say "For this Pants version, Pants requires Python ..."?

pants
```diff
@@ -73,23 +73,54 @@ EOF
...
+    if [[ "${pants_minor_version}" -ge 17 ]]; then
+      supported_python_versions=('3.6' '3.7' '3.8' '3')
+      supported_message='3.6+'
+    elif [[ "${pants_minor_version}" -eq 16 ]]; then
+      supported_python_versions=('3.6' '3.7' '2.7')
+      supported_message='2.7, 3.6, or 3.7'
+    elif [[ "${pants_minor_version}" -eq 15 ]]; then
+      supported_python_versions=('3.6' '2.7')
+      supported_message='2.7 or 3.6'
+    else
```
cosmicexplorer, line 107: Could we leave a comment here stating explicitly that the `else` case  is for `1.14.0` and below? It's not immediately clear without understanding the code.

cosmicexplorer (approved):
This is an inspired change which is thoughtfully implemented!

pants
cosmicexplorer: Perhaps leaving a TODO about `2.0.0` would be appropriate?

pants
```diff
@@ -73,23 +73,54 @@ EOF
...
+  pants_version="$(get_pants_ini_config_value 'pants_version')"
+  # If `pants_version` not pinned in `pants.ini`, default to what the most
+  # recent stable release supports.
+  if [[ -z "${pants_version}" ]]; then
+    supported_python_versions=('3.6' '2.7')
+    supported_message='2.7 or 3.6'
+  else
+    pants_minor_version="$(echo "${pants_version}" | cut -d '.' -f2)"
+    if [[ "${pants_minor_version}" -ge 17 ]]; then
+      supported_python_versions=('3.6' '3.7' '3.8' '3')
```
jsirois: This is not great. Ideally we'd just accept any python3.* And then use it to run a test snippet that checked `sys.version_info[:2] >= (3, 6) and sys.version_info[:2] < (4, 0)`. Then maybe sort the answers in the affirmative and pick the lowest. The main point being, do some slower stuff here to pick a 3 series interpreter if it avoids having to thrash this script again to support 3.9, etc. Currently we can't though since finding python is in the mainline of execution even after bootstrap :/. FWICT that is un-necessary. If we bootstrap pants version x.y.z, it's bootstrap path need not say which python if once the right choice for a given pants version, always the right choice. From my phone so I may not have taken in all the right context, but I think this makes sense.

pants
stuhood: I think that we know that `2.0.0` should definitely not use python `2.7`, which is the current effect of this code. I think it's a blocker, unfortunately. But it should only be two extra lines.

pants
Eric-Arellano: I'm a little confused how this would work. If there are multiple `python3`s when running `which -a python3`, how would we execute them all? Would we do something like this?

```
possible_pythons = ()
for bin in which -a python3
  version_info = sys.version_info[0:2]
  possible_pythons += (bin version_info)  # i.e. a tuple
done
find min from possible_pythons
```

Seems difficult to implement in Bash.

--

Alternatively, it's been stated somewhere that there won't be a 3.10 because Guido doesn't like two-digit minor versions. We could simply add `3.9` to the list and call it a day? Nothing bad will happen if `3.9` is not released yet, and we aren't claiming to support 4.

Is that fine? I can find the quote need be.

pants
Eric-Arellano: Done.

pants
```diff
@@ -73,23 +73,61 @@ EOF
...
+  # If `pants_version` not pinned in `pants.ini`, default to what the most
+  # recent stable release supports.
+  if [[ -z "${pants_version}" ]]; then
+    supported_python_versions=('3.6' '2.7')
+    supported_message='2.7 or 3.6'
+  else
+    pants_major_version="$(echo "${pants_version}" | cut -d '.' -f1)"
+    pants_minor_version="$(echo "${pants_version}" | cut -d '.' -f2)"
+    if [[ "${pants_major_version}" -eq 2 ]]; then
+      supported_python_versions=('3.6' '3.7' '3.8' '3.9')
```
Eric-Arellano, line 105: @stuhood should we make any assumptions about if we will support Python 3.6 in the 2.0 release? 

The main things we'd get by dropping Python 3.6 are context variables, `from __future__ import annotations` (useful for type hints) and improved `asyncio`. Python 3.6 is EOL in December 2021. I don't image we'd drop it until 2021 at the earliest for that reason.

stuhood: I don't think we should make any assumptions. Probably easiest to just have "everything greater than known version X" default to the max values for version X (ie `1.17.0` in this case).

Thank you!

pants
jsirois: I'm fine with handling 3.9 + quote. It would be nice to take interpreter selection out of the mainline still, but that could be a seperate change.

stuhood (approved):
Thank you Eric.

PR #50, opened 2019-04-02
Refactor .travis.yml by extracting install_python_for_ci.sh script

Eric-Arellano:
In https://github.com/pantsbuild/pants/pull/7470, we came up with a better design for interfacing with our git cloned version of Pyenv. By extracting out a script, we keep `.travis.yml` more focused and are able to improve the robustness of Pyenv-related code, such as adding comments.


cosmicexplorer (approved):
Thank you for taking the time to "backport" this change! I think this will make it absurdly easier to iterate on how we install python versions in the future.

Eric-Arellano:
Agreed! I think I mentioned to you how useful this setup repo has been as a place to iterate on how we do things in Pants repo. The entire caching of Pyenv started from this repo.

Excited to soon have Python 3 on all Pants CI shards so that we can start bringing over the scripts too.

PR #51, opened 2019-04-23
Dynamically determine the Pants version when not specified

Eric-Arellano:
### Problem
When following our install guide https://www.pantsbuild.org/install.html#recommended-installation, we end up bootstrapping *two* virtual environments and logging this to the user, which is confusing for your very first interaction with Pants.

This is because the script names the venv folder as `unspecified_py*` when `pants.ini` does not have a version pinned: https://github.com/pantsbuild/setup/blob/8fca2f4a51ddc65c931d4aba4ec3fbc1acf27234/pants#L154-L158.

### Solution
PyPi has a JSON API that we can ping to determine dynamically what the most recent stable release is. This is the same version `pip` will resolve to, so we simply name the venv folder this.

This also allows us to change our execution flow, so that we first determine the Pants version and _then_ determine the Python version, rather than the reverse. The benefit here is that the Python function solely depends on the Pants version, as of https://github.com/pantsbuild/setup/pull/49, and now the execution flow reflects this properly.

#### Testing
We add tests to ensure this all works.

In doing so, we add a new infrastructure for hermitic tests. Rather than acting directly on this repo's `pants.ini`, we use temporary directories to create a new, clean environment every run. This infrastructure will be used to refactor and improve the rest of this repo's tests in followup PRs.

### Result
If the `pants_version` is left off of `pants.ini`, Pants will setup the venv folder to use the most recent stable release.

### DIFF
--- a/build-support/test_first_time_install.py
+++ b/build-support/test_first_time_install.py
@@ -1,0 +1,62 @@
+#!/usr/bin/env python3
+# Copyright 2019 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+"""Test that the first time installation flow described by
+https://www.pantsbuild.org/install.html#recommended-installation works as expected.
+
+Note that this does not test the result of `./pants generate-pants-ini`, because that
+is already tested by Pants at
+https://github.com/pantsbuild/pants/blob/master/tests/python/pants_test/core_tasks/test_generate_pants_ini.py."""
+
+import subprocess
+import unittest
+
+from common import setup_pants_in_tmpdir, travis_section
+
+
+class TestFirstTimeInstall(unittest.TestCase):
+
+  def test_venv_name_uses_most_recent_stable_release(self) -> None:
+    with setup_pants_in_tmpdir() as tmpdir:
+      completed_process = subprocess.run(
+        ["./pants", "--version"],
+        check=True,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        encoding="utf-8",
+        cwd=tmpdir
+      )
+      # Pip will resolve the most recent stable release, which will be the output of
+      # `./pants --version`.
+      downloaded_version = completed_process.stdout[-2]
+      virtual_env_created_log_entry = next(
+        line for line in completed_process.stderr.split("\n")
+        if "created" in line and "bootstrap" in line
+      )
+      self.assertIn(downloaded_version, virtual_env_created_log_entry)
+
+  def test_only_bootstraps_the_first_time(self) -> None:
+    with setup_pants_in_tmpdir() as tmpdir:
+      first_run_pants_script_logging = subprocess.run(
+        ["./pants", "--version"],
+        check=True,
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.PIPE,
+        encoding="utf-8",
+        cwd=tmpdir
+      ).stderr
+      self.assertTrue(first_run_pants_script_logging)
+      second_run_pants_script_logging = subprocess.run(
+        ["./pants", "--version"],
+        check=True,
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.PIPE,
+        encoding="utf-8",
+        cwd=tmpdir
+      ).stderr
+      self.assertFalse(second_run_pants_script_logging)
+
+if __name__ == "__main__":
+  with travis_section("PantsFirstTimeInstall", "Testing first time install."):
+    unittest.main()
--- a/build-support/test_first_time_install.py
+++ b/build-support/test_first_time_install.py
@@ -9,6 +9,7 @@ Note that this does not test the result of `./pants generate-pants-ini`, because
 is already tested by Pants at
 https://github.com/pantsbuild/pants/blob/master/tests/python/pants_test/core_tasks/test_generate_pants_ini.py."""
 
+import re
 import subprocess
 import unittest
 
@@ -32,7 +33,7 @@ class TestFirstTimeInstall(unittest.TestCase):
       downloaded_version = completed_process.stdout[-2]
       virtual_env_created_log_entry = next(
         line for line in completed_process.stderr.split("\n")
-        if "created" in line and "bootstrap" in line
+        if re.search(r"virtual environment successfully created at .*/bootstrap/.*_py", line)
       )
       self.assertIn(downloaded_version, virtual_env_created_log_entry)
 
--- a/build-support/test_first_time_install.py
+++ b/build-support/test_first_time_install.py
@@ -46,6 +46,7 @@ class TestFirstTimeInstall(unittest.TestCase):
         encoding="utf-8",
         cwd=tmpdir
       ).stderr
+      # Check that stderr is truthy, i.e. there is output.
       self.assertTrue(first_run_pants_script_logging)
       second_run_pants_script_logging = subprocess.run(
         ["./pants", "--version"],
@@ -55,6 +56,7 @@ class TestFirstTimeInstall(unittest.TestCase):
         encoding="utf-8",
         cwd=tmpdir
       ).stderr
+      # Check that stderr is falsy, i.e. there is no output.
       self.assertFalse(second_run_pants_script_logging)
 
 if __name__ == "__main__":
--- a/build-support/test_first_time_install.py
+++ b/build-support/test_first_time_install.py
@@ -30,12 +30,11 @@ class TestFirstTimeInstall(unittest.TestCase):
       )
       # Pip will resolve the most recent stable release, which will be the output of
       # `./pants --version`.
-      downloaded_version = completed_process.stdout[-2]
-      virtual_env_created_log_entry = next(
-        line for line in completed_process.stderr.split("\n")
-        if re.search(r"virtual environment successfully created at .*/bootstrap/.*_py", line)
-      )
-      self.assertIn(downloaded_version, virtual_env_created_log_entry)
+      downloaded_version = completed_process.stdout.strip()
+      self.assertTrue(re.search(
+        fr"virtual environment successfully created at .*/bootstrap.*/{downloaded_version}_py",
+        completed_process.stderr, flags=re.MULTILINE
+      ))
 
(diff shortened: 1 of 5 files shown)


jsirois (approved):

pants
```diff
@@ -72,42 +72,46 @@ EOF
...
+  else
+    curl -sSL https://pypi.python.org/pypi/pantsbuild.pants/json |
+      python -c "import json, sys; print(json.load(sys.stdin)['info']['version'])"
+  fi
+}
+
+function determine_default_python_exe {
+  pants_version="$1"
+  pants_major_version="$(echo "${pants_version}" | cut -d '.' -f1)"
+  pants_minor_version="$(echo "${pants_version}" | cut -d '.' -f2)"
```
cosmicexplorer, line 101: This could be extracted into a function, like:
```bash
function extract_version_field {
  version_string="$1"
  field_index="$2"
  echo "$version_string" | cut -d '.' "-f${field_index}"
}
```

It's not clear that's necessary right now and I believe others would probably argue against it right now, but I think for a well-defined operation like this it could help to add clarity.

pants
```diff
@@ -72,42 +72,46 @@ EOF
...
 # The high-level flow:
-# 1.) Resolve the Python interpreter, first reading from the env var $PYTHON,
+# 1.) Resolve the Pants version from pants.ini or from the latest stable
+#     release to PyPi.
+# 2.) Resolve the Python interpreter, first reading from the env var $PYTHON,
 #     then reading from pants.ini, then determining the default based off of
-#     which Python versions the current `pants_version` supports.
-# 2.) Grab pants version from pants.ini or default to latest.
+#     which Python versions the Pants version supports.
 # 3.) Check for a venv via a naming/path convention and execute if found.
```
cosmicexplorer: Could we maybe add a really quick description of why we need to resolve the pants version first now (I guess not just because we're using it resolve the python interpreter now, but also that it's mixed into the venv dir location)?

cosmicexplorer (commented):
> This edge case is covered currently by CI via the `./build-support/ci.py --pants-version unspecified` tests.

Does that currently test that we're not bootstrapping more than once?

pants
Eric-Arellano: If we ever need to parse a field other than `pants_version`, I agree. Otherwise this for now seems to be making things more general and abstract than necessary. 

I think the current implementation is concise and readable - even if you don't know how `cut` works, it's easy to get the gist.

pants
Eric-Arellano: Good idea! Fixed, along with updating some other parts of the docs so that we better explain the flow and fix things like link to the site.

Eric-Arellano (commented):
> Does that currently test that we're not bootstrapping more than once?

No, it does not. `ci.py` currently does not inspect any of the actual venv folder code, in large part due to concerns about hermicity. To ensure consistent CI runs, we would need to start purging the CI user's Pants `.cache/pants/` entries, which I'm not sure is a great idea.

Confirmed this works by testing locally though. And Travis logging shows that it correctly names the folder as the version, rather than `unspecified`.

cosmicexplorer:
> Confirmed this works by testing locally though. And Travis logging shows that it correctly names the folder as the version, rather than unspecified.

Can we test this?

Eric-Arellano:
>> Confirmed this works by testing locally though. And Travis logging shows that it correctly names the folder as the version, rather than unspecified.
>
> Can we test this?

I'm not sure how to test this properly due to the below.

> ci.py currently does not inspect any of the actual venv folder code, in large part due to concerns about hermicity. To ensure consistent CI runs, we would need to start purging the CI user's Pants .cache/pants/ entries, which I'm not sure is a great idea.

--

I think we probably should add this type of testing? Our testing is not really testing the implementation of this script at the moment, in multiple ways. All it tests is that given a certain combination of `pants_version` and `pants_runtime_python_version` settings, the script is able to eventually run `./pants -V` and `./pants list ::`.

We are not testing these things currently:
- the venv folder has the correct name.
- `./pants -V` is the version we actually wanted.

Then once we add this PR, add that the venv folder is not created again when running `./pants` after having ran `./pants generate-pants-ini`.

If we decide we do want to start testing that the venv folder is created properly, then I think this will need to be broken out as a precursor PR.

--

Actually one idea for hermitic tests... Override the cache location to something custom so that we can assume it's empty and then can delete the folder after each test.

https://github.com/pantsbuild/setup/blob/9aa4a34d032493cd6e198e42a0de11b4201698c2/pants#L18

How does this all sound? Thinking out loud here.

One concern I have is coupling the tests too much to the bash script. Should `ci.py` be coupled to the naming scheme for the venv? I think so, but worth confirming.

cosmicexplorer:
I was more referring to:
> And Travis logging shows that it correctly names the folder as the version, rather than unspecified.

If travis logging shows that the folder is correctly named, we could read the logging output and verify this perhaps?

build-support/test_first_time_install.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python3
```
Eric-Arellano, line 1: We now swallow the output of the Pants script and of `./pants --version` in these tests. Do you all think that's a good thing, or should we add a line to `print(completed_process.stderr)`? 

Note that if the tests fail, then we will output the relevant parts. This is only asking of the value of logging every time so that people can in Travis see what `./pants` ends up logging to the user.

Eric-Arellano: After working on followup PRs that build off of this new testing infrastructure—namely converting `ci.py` into `test_sanity_check.py`—I think we should continue to swallow the output *but* always ensure we print the output somehow upon failures. This PR currently does that, so I think we should keep as is.

Eric-Arellano: This question is a bit irrelevant actually.

The next followup PR is to convert this test into a Pants target, meaning it will be run through Pytest. Pytest by default silences all stdout and stderr, and only logs if the test fails (https://docs.pytest.org/en/latest/capture.html). This sounds like the ideal behavior. So, in the followup, we can remove the `subprocess.DEVNULL` lines and let the script log normally.

build-support/test_first_time_install.py
```diff
@@ -0,0 +1,62 @@
...
+        stderr=subprocess.PIPE,
+        encoding="utf-8",
+        cwd=tmpdir
+      )
+      # Pip will resolve the most recent stable release, which will be the output of
+      # `./pants --version`.
+      downloaded_version = completed_process.stdout[-2]
+      virtual_env_created_log_entry = next(
+        line for line in completed_process.stderr.split("\n")
+        if "created" in line and "bootstrap" in line
```
cosmicexplorer: I would prefer that this line use a regex match, i.e.:
```python
re.search(r'^.*created.*bootstrap.*/path/to/bootstrapdir/{}.*$'.format(downloaded_version),
          line, flags=re.MULTILINE)
```

build-support/test_first_time_install.py
```diff
@@ -0,0 +1,62 @@
...
+      self.assertTrue(first_run_pants_script_logging)
+      second_run_pants_script_logging = subprocess.run(
+        ["./pants", "--version"],
+        check=True,
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.PIPE,
+        encoding="utf-8",
+        cwd=tmpdir
+      ).stderr
+      self.assertFalse(second_run_pants_script_logging)
```
cosmicexplorer, line 60: Could we add a line explaining that we're checking that the stderr is empty? It isn't immediately clear from reading the code (perhaps it's just super pythonic).

pants
```diff
@@ -1,14 +1,14 @@
 #!/usr/bin/env bash
-# Copyright 2015 Pants project contributors (see CONTRIBUTORS.md).
```
cosmicexplorer: I do not think we should be changing the copyright year here. Am I missing the reason for this change?

pants
```diff
@@ -1,14 +1,14 @@
...
 # =============================== NOTE ===============================
-# This pants bootstrap script comes from the pantsbuild/setup
+# This ./pants bootstrap script comes from the pantsbuild/setup
 # project and is intended to be checked into your code repository so
 # that any developer can check out your code and be building it with
-# pants with no prior setup needed.
+# Pants with no prior setup needed.
 #
-# You can learn more here: https://pantsbuild.github.io/setup
+# You can learn more here: https://www.pantsbuild.org/install.html
```
cosmicexplorer, line 11: Great fix.

build-support/test_first_time_install.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python3
+# Copyright 2019 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+"""Test that the first time installation flow described by
+https://www.pantsbuild.org/install.html#recommended-installation works as expected.
+
+Note that this does not test the result of `./pants generate-pants-ini`, because that
+is already tested by Pants at
+https://github.com/pantsbuild/pants/blob/master/tests/python/pants_test/core_tasks/test_generate_pants_ini.py."""
```
cosmicexplorer, line 10: Good docstring!

cosmicexplorer (changes requested):
Only marking as "request changes" because I'm not sure we should be changing the copyright year. Thanks for cleaning up the urls/etc in the script!

pants
Eric-Arellano: I know usually we shouldn't update copyright, but I wanted to make an exception here so that the script does not seem outdated, since users interact so directly with this one.

From what I can tell, the main reason we don't update copyright year is to act as a historical record, right? In Pants code, which users don't directly interact with, I think that's great. Here, that is less of a priority in my opinion.

Can revert back if this is still wrong to make the exception.

cosmicexplorer: cc @benjyw who had thoughts on this the last time I remember talking about this^

build-support/test_first_time_install.py
cosmicexplorer: +1 to the conclusion here!

cosmicexplorer (approved):
Great modulo the copyright year, which may be fine! If we can move that discussion to a followup PR, let's definitely merge this :D Thanks!!

PR #52, opened 2019-04-23
Test Python 37 support

Eric-Arellano:
We restored support in https://github.com/pantsbuild/pants/pull/7578. Now that 1.16.0.dev2 is out, we should confirm this works before the stable release is cut.

We must avoid running these new tests on Ubuntu Precise and Trusty because they do not have Python 3.7 installed.

We also refactor how we handle `.travis.yml` script entries + allow skipping Pantsd tests via the env var `SKIP_PANTSD_TESTS` to result in less duplication.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,12 +14,17 @@ env:
     - PATH="${PYENV_ROOT}/shims:${PATH}"
 
 script:
-  - ./build-support/ci.py
+  - &test_1_14 >
+    ./build-support/ci.py
     --pants-versions 1.14.0
     --python-versions unspecified
-  - ./build-support/ci.py
+  - &test_1_15_and_unspecified >
+    ./build-support/ci.py
     --pants-versions unspecified 1.15.0
     --python-versions unspecified 2.7 3.6
+  - ./build-support/ci.py
+    --pants-versions config
+    --python-versions unspecified 2.7 3.6 3.7
 
 cache:
   directories:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -14,12 +14,17 @@ env:
     - PATH="${PYENV_ROOT}/shims:${PATH}"
 
 script:
```
jsirois, line 41: This is confusing since - fwict it is unused except as an holder for anchors. If I'm right in that characterization, consider `x-script:` or `shared-scripts:` or some other non-overloaded name.

jsirois: I'm not right in that characterization - several matrix entries do inherit these scripts :/

I am a fan of:
```yaml
      script:
        - *test_1_14
        - *test_1_15_and_unspecified
```

Consider doing more of those and being explicit with the scripts for every matrix entry. This would have all scripts definied in one place up here - in x-scripts say - and referenced below explicitly in each job. Maybe. I just find these complex yaml configs very confusing to read. It may just be the nature of the beast.

jsirois (approved):

.travis.yml
Eric-Arellano: Good idea. I refactored the whole thing to take a similar approach, which has the benefit of making it much more explicit where we skip Py37 and will make it easier to update these entries as we release more stable versions.

Thanks for the suggestion.

jsirois (approved):

PR #53, opened 2019-04-25
Run `shellcheck.py` over build-support scripts

Eric-Arellano:
Test code should be high quality too. We can cheaply improve our build-support code quality by extending `shellcheck.py` to run against it, in addition to checking `./pants`.


blorente (approved):
Thanks! I learned a bunch of bash syntax from this.

PR #54, opened 2019-04-25
Mark scripts and tests as Pants targets to unlock benefits like using Pytest and isort

Eric-Arellano:
Because our tests and build-support scripts are written in Python, they are valid candidates for Pants targets.

Marking them as targets allows for the `build-support` and `tests` code to improve in multiple ways:

* Can use Pytest via `./pants test ::`, which brings benefits like capturing stdout and stderr
* Can run `./pants fmt.isort ::` for import checks
* Can run `./pants lint ::` to catch bugs in tests and scripts
* Can run `./pants mypy ::` to catch bugs in tests and scripts
* Allows creating / importing new `TestBase`, so that test support code can be kept separate from script support code (i.e. `common.py`)

### DIFF
--- a/pants.ini
+++ b/pants.ini
@@ -4,3 +4,6 @@ pants_version: 1.16.0.dev2
 plugins: [
     'pantsbuild.pants.contrib.go==%(pants_version)s',
   ]
+
+[python-setup]
+interpreter_constraints: ['CPython>=3.6']
--- a/pants.ini
+++ b/pants.ini
@@ -1,5 +1,5 @@
 [GLOBAL]
-pants_version: 1.16.0.dev2
+pants_version: 1.16.0.dev3
 
 plugins: [
     'pantsbuild.pants.contrib.go==%(pants_version)s',
--- a/pants.ini
+++ b/pants.ini
@@ -6,4 +6,4 @@ plugins: [
   ]
 
 [python-setup]
-interpreter_constraints: ['CPython>=3.6']
+interpreter_constraints: ['CPython>=3.6,<4']
--- a/tests/BUILD
+++ b/tests/BUILD
@@ -1,0 +1,15 @@
+# Copyright 2019 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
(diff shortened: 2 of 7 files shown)


pants.ini
```diff
@@ -1,6 +1,9 @@
 [GLOBAL]
-pants_version: 1.16.0.dev2
+pants_version: 1.16.0.dev3
 
 plugins: [
     'pantsbuild.pants.contrib.go==%(pants_version)s',
   ]
+
+[python-setup]
+interpreter_constraints: ['CPython>=3.6']
```
stuhood: Would recommend setting a max here, to minimize the chance of being surprised in the future. 4?

tests/BUILD
```diff
@@ -0,0 +1,15 @@
+# Copyright 2019 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+python_library(
+  name = 'test_base',
+  sources = 'test_base.py',
+)
+
```
stuhood, line 8: To get a useful package/module name for code, it's recommended to locate it below a source root: https://www.pantsbuild.org/setup_repo.html#source-roots

stuhood (approved):
Would recommend moving the code under a source root, but not really a blocker.

Thanks!

tests/BUILD
Eric-Arellano: I don't see how to do this while keeping the flat folder. Changing it to `tests/python/tests` gets the import `tests.test_base` working, but I would love to not have to do this because it means a longer command every time (`./pants test tests/python/tests:first_time_install` vs. `./pants test tests:first_time_install`) and this repo will never be multilingual.

I tried playing around with the options from `./pants help-advanced source`, but was confused what they mean.

Is there a way to do this without changing the folder structure at all?

Eric-Arellano:
> Would recommend moving the code under a source root

Does this mean move `./pants` into `src` or your recommendation on making `tests` have a source root?

Eric-Arellano:
Merging now, as this blocks several followup PRs. Happy to address the source roots suggestion in a followup PR though!

PR #55, opened 2019-04-29
Pull/54 debug

jsirois:


.travis.yml
```diff
@@ -14,27 +14,22 @@ env:
...
-  ./build-support/ci.py
-  --pants-versions config
-  --python-versions unspecified 2.7 3.6
-
 test_first_time_install: &test_first_time_install >
-  ./build-support/test_first_time_install.py
+  PEX_VERBOSE=9 ./pants test tests:first_time_install
 
 script:
   - *test_1_14
```
Eric-Arellano: The `ci.py` files should pass perfectly fine. I recommend not running those scripts to save CI time (about 6 minutes).

The only thing that fails is `*test_first_time_install`, because it's the only thing to trigger the `select-interpreter` task thanks to running `./pants test` with constraints defined in `pants.ini`.

jsirois: Yup, figured that out a bit too late. Thanks.
Nonetheless, the key data is now secured.

jsirois:
This answered question of what was up so I'll close.

PR #56, opened 2019-04-29
Stop using Pyenv shims for custom installed Python

Eric-Arellano:
Shims are designed for humans, not machines. Most tangibly, they resulted in this unexpected behavior https://github.com/pantsbuild/pants/issues/7636 on the OSX 10.12 shard.

We still use `pyenv global` on the Precise and Xenial shards, however, because Travis pre-installs Python via Pyenv on those shards and this is the standard way to fix an issue with their implementation so that the interpreters become discoverable on the `PATH`. We stick with Travis's pre-installed versions for performance and a smaller cache, at the cost of a less uniform and more complex Python setup across CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,8 @@ env:
     # NB: Linux shards use Pyenv to pre-install Python. We must not override
     # PYENV_ROOT on those shards, or their Python will no longer work.
     - PYENV_ROOT="${PYENV_ROOT:-${HOME}/.pants_pyenv}"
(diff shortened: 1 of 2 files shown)


jsirois:
> We still use pyenv global on the Precise and Xenial shards...

It might be worth noting further that we only engage this hack for speed purposes; otherwise we could use a standard pyenv setup routine on osx and linux. That is, IIUC.

jsirois (approved):

PR #58, opened 2019-04-30
Add `precommit.py` script to easily lint all the repo's code

Eric-Arellano:
Provides one entry point to check the import order, Python lint, MyPy types, and to run shellcheck.

This is not yet converted into a formal git hook to avoid adding too much complexity.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 /.pants.workdir.file_lock
 /.pids/
 
+/.cache/
 /.mypy_cache/
 
 __pycache__/
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,7 +79,7 @@ linux_setup: &linux_setup
 
 matrix:
   include:
-    - name: "Lint ./pants script"
+    - name: "Lint scripts, tests, and build-support"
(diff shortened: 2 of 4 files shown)


blorente (approved):
Thanks! I would very much be okay with this being a pre-push (or pre-commit) hook.

Eric-Arellano:
> I would very much be okay with this being a pre-push (or pre-commit) hook.

Me too, I'd love for it to be a pre-commit hook. But I didn't spend the time to see how feasible this would be to do. I'll check today before merging this. I would also love Pex to have a pre-commit check for `tox -v -e style`, so doing the research now could help with that too.

Eric-Arellano:
Eh it appears we would need to make it easy for developers to have a file named `.git/hooks/pre-commit` in their local repo, which notably would not be tracked by VCS. This is why in Pants we have the script [`install_git_hooks.sh`](https://github.com/pantsbuild/pants/blob/master/build-support/bin/install_git_hooks.sh). 

Setting up our repo or Pex with this script is getting a bit more complex than I'd like.

PR #59, opened 2019-04-30
Add MyPy config and script

Eric-Arellano:
Since most of the code in this repo now uses Python 3 with type hints, it would be good to actually enforce that those hints are correct in CI.

This PR also allows us to experiment with MyPy to upstream to Pants, e.g. copying the new Python script over.

### DIFF
--- a/build-support/ci.py
+++ b/build-support/ci.py
@@ -8,7 +8,7 @@ import os
 import subprocess
 from contextlib import contextmanager
 from enum import Enum
-from typing import List
+from typing import Iterator, List
 
 from common import (CONFIG_GLOBAL_SECTION, banner, die, read_config,
                     temporarily_rewrite_config, travis_section)
@@ -27,8 +27,8 @@ class PantsVersion(Enum):
   one_sixteen = "1.16.0"
   one_seventeen = "1.17.0"
 
-  def __str__(self):
-      return self.value
+  def __str__(self) -> str:
+      return str(self.value)
 
 
 class PythonVersion(Enum):
@@ -37,8 +37,8 @@ class PythonVersion(Enum):
   py36 = "3.6"
   py37 = "3.7"
 
-  def __str__(self):
-    return self.value
+  def __str__(self) -> str:
+    return str(self.value)
 
 
 def main() -> None:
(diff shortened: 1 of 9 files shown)


tests/test_base.py
```diff
@@ -21,16 +22,17 @@ def copy_pants_into_tmpdir(self):
...
   @contextmanager
-  def set_pants_cache_to_tmpdir(self):
+  def set_pants_cache_to_tmpdir(self) -> Iterator[None]:
     with tempfile.TemporaryDirectory() as tmpdir:
       original_env = os.environ.copy()
       os.environ["PANTS_HOME"] = tmpdir
       try:
         yield
       finally:
-        os.environ = original_env
```
Eric-Arellano, line 31: Bug caught by MyPy!

pants.ini
```diff
@@ -2,7 +2,7 @@
 pants_version: 1.16.0.dev3
 
 plugins: [
-    'pantsbuild.pants.contrib.go==%(pants_version)s',
```
blorente, line 5: Do we not need `go` anymore? If so, maybe worth removing in a separate PR. This is a nit, feel free to ignore it :)

build-support/ci.py
```diff
@@ -37,8 +37,8 @@ class PythonVersion(Enum):
   py36 = "3.6"
   py37 = "3.7"
 
-  def __str__(self):
-    return self.value
+  def __str__(self) -> str:
```
blorente, line 40: Another potential bug caught? Nice!

blorente (approved):
This is so cool! thanks!

pants.ini
Eric-Arellano: No, from what I can tell we were solely using `go` as a sanity check that any contrib package worked in general. From what I can tell, there was no particular reason to choose `go` and `mypy` will get the same job done, while allowing us to also type check our code.

build-support/ci.py
Eric-Arellano: Yes! Didn't fix any actual bug in our code, but a potential one yes.

What's going on is that an `Enum`'s value is type `Any`. Above, we always are setting it to `str`, but you could just as well do something like `py36 = 3.6` (i.e. `float`). MyPy doesn't infer that we are always using `str`, because there is 0 guarantee that we will continue to do this. It is totally possible in Python to set one enum value as `str` and the next to `float`. So, this fix is a bit conservative but technically is correct to do.

PR #61, opened 2019-10-23
Add fixed locale settings

bgokden:
LC_ALL=en_US.UTF-8
 LANG=en_US.UTF-8

are needed for pants to work.

pants create this error:

```shell
pants.bin.pants_loader.InvalidLocaleError: 
Your system's preferred encoding is `US-ASCII`, but Pants requires `UTF-8`.
Specifically, Python's `locale.getpreferredencoding()` must resolve to `UTF-8`.

Fix it by setting the LC_* and LANG environment settings. Example:
  LC_ALL=en_US.UTF-8
  LANG=en_US.UTF-8
Or, bypass it by setting the below environment variable. 
  PANTS_IGNORE_UNRECOGNIZED_ENCODING=1
Note: we cannot guarantee consistent behavior with this bypass enabled.
```

so it is better to set it to a fixed value in the setup script.
Pants create its own virtual env so the it doesn't read the system settings on MacOS.

### DIFF
--- a/pants
+++ b/pants
@@ -190,4 +190,4 @@ pants_dir="$(bootstrap_pants "${pants_version}" "${python}")"
 # See https://blog.phusion.nl/2017/10/13/why-ruby-app-servers-break-on-macos-high-sierra-and-what-can-be-done-about-it/
 export no_proxy='*'
 
-exec "${pants_dir}/bin/python" "${pants_dir}/bin/pants" "$@"
+exec env LC_ALL=en_US.UTF-8 env LANG=en_US.UTF-8 "${pants_dir}/bin/python" "${pants_dir}/bin/pants" "$@"


Eric-Arellano (commented):
Thanks for this patch @bgokden! 

I'm a bit conflicted. I know this solves a problem for some folks, but I don't love the idea of hardcoding this for every user by default because that is surprising to do. Instead, I think it's better for users to modify their copy of `./pants` to have this override as you do here. Thoughts @jsirois?

jsirois:
> I'm a bit conflicted. I know this solves a problem for some folks, but I don't love the idea of hardcoding this for every user by default because that is surprising to do.

I'm not sure what we gain by having folks do this manually since Pants requires it to work reasonably. We already do the ~inverse for example and scrub environment variables we don't want to see without consulting the user.

That said, I don't think things are as simple as exporting the env vars in line. That assumes the encodings exported actually exist. I know on my arch machine I need to generate the locales I want support for: https://wiki.archlinux.org/index.php/locale#Generating_locales.

So, I'd say I'm in favor of this change iff the error you get when `en_US.UTF-8` is not present on your machine is still useful, actionable like the current error message is.

jsirois:
Ew, and all that said, do we really require the en_US bit? That seems unlikely and a bit horrible.

Eric-Arellano:
> Ew, and all that said, do we really require the en_US bit? That seems unlikely and a bit horrible.

Negative, we require UTF-8 and in our help message give an example of using `en_US`. Possibly, it should be reworded to clarify all that matters is the UTF 8 piece.

https://github.com/pantsbuild/pants/blob/80e19dcd14c569faa5d2550c20ce169b4704aa49/src/python/pants/bin/pants_loader.py#L44-L57

jsirois:
OK - thanks for doing my homework for me. That alone seems to suggest to me this fix is not a great idea as written.

bgokden:
@jsirois This one worked for me but I am not sure if it works for everyone. I think a better solution should be written, but we are not using pants anymore so I am not working on this anymore.

Thanks for the feedback. I missed your previous comment.

PR #63, opened 2019-12-02
Update copyright to current year

jsoref:


stuhood:
Hey @jsoref : we don't currently have a policy of editing copyright headers on otherwise idle files. Do you know of a reason why we should?

benjyw:
Generally speaking we want copyright headers to reflect the earliest year for which we claim copyright, so this change seems wrong to me.

jsoref:
Generally copyright should reflect the latest year with a meaningful change. It's also useful to have it reflect the first year.

The reason I'm asking for just this file is that when we looked at the pants build grabbing script, the only date it showed is this old one. That made it look like the file/project was unhealthy.

I can get you a reference for how others handle this later.

PR #64, opened 2020-01-13
Use Tox instead of `build-support` scripts

Eric-Arellano:
When we expanded our CI in May 2019, we took the approach of Pants of having multiple lint utility scripts in `build-support`, rather than Pex's approach of using `tox`. This was mostly due to my unfamiliarity with Tox.

Beyond unnecessary custom code, another downside to this `build-support` approach is that it poses a chicken-and-the-egg problem when developing this repo. For example, for our tests and linters to run properly, `./pants` must work; however, the entire point of our linters and tests is to confirm that the script `./pants` works in the first place. Using `tox` allows us to run our tests and lints regardless of the condition of the `./pants` script.

This also does the following:
* Adds to CI macOS Mojave and Ubuntu 18.04.
* Applies Black formatting
* Uses Flake8 rather than the custom Pants checker


README.md
```diff
@@ -1,3 +1,21 @@
...
+
+### Development
+
+We use [tox](https://testrun.org/tox/en/latest/) for test automation. To run the test suite and linters, invoke tox:
+
+```bash
+$ tox
+```
+
+You may run certain environments with `tox -e` (run `tox -a` to see all options):
```
blorente:
```suggestion
You may run certain checks (`environments` in tox) with `tox -e` (run `tox -a` to see all options):
```

The fact that they call them `environments` but they actually _run_ the thing was a super big gotcha for me when I touched tox for the first time.

build-support/common.sh
```diff
@@ -15,43 +9,3 @@ function die() {
   (($# > 0)) && log "\n${COLOR_RED}$*${COLOR_RESET}"
   exit 1
 }
-
```
blorente, line 18: Good job on getting rid of tons of code!

blorente (approved):
I deeply dislike tox, but I think it's because of my lack of understanding. In principle however, I think these are good changes, particularly because of the chicken-and-egg reason you mentioned.

It would be good to include if at all possible how this affects the look of CI, maybe with a couple screenshots.

README.md
Eric-Arellano: Good suggestion!

FWIT, I think the word `environments` comes from the fact that one of the main features of Tox is to allow testing your library with different versions of Python, e.g. `py27` vs. `py36`. These are distinct "environments". 

Still, confusing for first-time users and I like how your change uses familiar language while still clarifying what the official Tox verbiage is.

Eric-Arellano (commented):
> I deeply dislike tox, but I think it's because of my lack of understanding.

I felt the same way for a while and tox intimidated me when first working on Pex (my first introduction to Tox). I've grown to like it much more now. 

One of the main things that helped was once I learned the 2-3 commands I needed for my typical workflow, such as `tox -e format-run`. I don't think this was in the Pex README originally and I found the Tox docs a bit confusing, so I went some time thinking "Okay, but how do I run isort?" We changed the Pex README to now explain that. This repo's README rips most of Pex's explanation. I tried to strike a balance between over-explaining vs. giving just enough to get running with the tool.

> It would be good to include if at all possible how this affects the look of CI, maybe with a couple screenshots.

I'm not quite sure what you mean? CI output is almost the same as before outside of slightly different logging.

.travis.yml
```diff
@@ -51,9 +54,9 @@ linux_setup: &linux_setup
     - "3.6"
     - "3.7"
```
jsirois, line 55: Looking at the output of all shards, in each case, exactly 1 python interpreter is used - 3.7.6 for OSXs, 3.6.* for linuxes. As such, a follow up would be good to kill all the extra python setup including lines like `pyenv global 2.7.17 3.6.9 3.7.5 3.8.0` and `./build-support/install_python_for_ci.sh "${PYENV_PY36_VERSION}" "${PYENV_PY37_VERSION}"` which would seem to be either un-needed or doing too much.

.travis.yml
```diff
@@ -90,6 +93,13 @@ matrix:
         - *env_osx_openssl
```
jsirois, line 93: `SKIP_PANTSD_TESTS=true` above in the fold is now cruft.

jsirois (commented):
This looks good, but there does seem to be both python version and pants cruft still laying about. A second pass would be good.

.travis.yml
Eric-Arellano: Negative, we are still testing all the different Python versions within the test itself:

https://github.com/pantsbuild/setup/blob/000271094eea1bc3b00dbde8646a4401e09b6af2/tests/test_sanity_check.py#L44-L50

.travis.yml
Eric-Arellano: Negative. We are still testing pantsd:

https://github.com/pantsbuild/setup/blob/000271094eea1bc3b00dbde8646a4401e09b6af2/tests/test_sanity_check.py#L30-L33

.travis.yml
Eric-Arellano: Example of how we are still running Python 2.7: the most recent CI run failed because of Python 2.7 installs. See https://github.com/pantsbuild/setup/pull/65.

blorente:
This was merged into the gh-pages branch. Was that intentional?

Eric-Arellano:
> This was merged into the gh-pages branch. Was that intentional?

Yes, this repository's ~master branch is called `gh-pages` so that we can point the script at a nice URL for people to download.

PR #66, opened 2020-02-22
Fix disabled Pyenv versions breaking the script

### LINKED ISSUE #62: pants 1.21.0 setup failure when python 3.7 is installed but disabled via pyenv
Pants version: 1.21.0
Pants installed via: `curl -L -O https://pantsbuild.github.io/setup/pants && chmod +x pants
`

Update the `pants.ini` to `pants_version: 1.21.0` then run `./pants`

```
C02XV41NJHD2:mlp-ark zhanwu.xiong$ ./pants
pyenv: python3.7: command not found

The `python3.7' command exists in these Python versions:
  3.7.4

pyenv: python3.7: command not found


The `python3.7' command exists in these Python versions:
  3.7.4

./pants: line 170: /Users/zhanwu.xiong/.cache/pants/setup/bootstrap-Darwin-x86_64/pants.35pSjk/install/bin/pip: No such file or directory
./pants: line 171: /Users/zhanwu.xiong/.cache/pants/setup/bootstrap-Darwin-x86_64/pants.35pSjk/install/bin/pip: No such file or directory
New virtual environment successfully created at /Users/zhanwu.xiong/.cache/pants/setup/bootstrap-Darwin-x86_64/1.21.0_py.
./pants: line 193: /Users/zhanwu.xiong/.cache/pants/setup/bootstrap-Darwin-x86_64/1.21.0_py/bin/python: No such file or directory
```

Full `pants.ini` after update:
```
[GLOBAL]
pants_version: 1.21.0
print_exception_stacktrace: True

pythonpath: ["%(buildroot)s/plugins"]
plugins: [
    'pantsbuild.pants.contrib.go==%(pants_version)s',
    'pantsbuild.pants.contrib.python.checks==%(pants_version)s',
    'docker==4.0.2']

backend_packages : [
    'pants.backend.graph_info',
    'pants.backend.python',
    'pants.backend.native',
    'pants.backend.project_info',
    'pants.backend.codegen.grpcio.python',
    'pants.rules.core',
    'cruise.docker']

[python-setup]
interpreter_constraints: ["CPython>=3.6"]

[source]
source_root_patterns = ['src/python/*', 'src/*', 'src/main/*', 'src/scala/*', 'src/java/*']
# Per language reverse key map.
# Go lang is designed to be built from Monorepo structure
source_roots = {u'go/src': (u'go',), u'go/main/src': (u'go',)}

[python-repos]
indexes = ['https://artifactory.robot.car/artifactory/api/pypi/wheels-python-local/simple', 'https://pypi.org/simple/']
repos = []

[repl.py]
ipython = True

[go-distribution]
version: 1.12

[lint.python-eval]
skip: True
```

The immediate cause of the issue is pants setup believes `python3.7` is a valid python version, but actually it is not. See the demonstration below:

```
C02XV41NJHD2:mlp-ark zhanwu.xiong$ pyenv versions
  system
  2.7.15
* 3.6.9 (set by /Users/zhanwu.xiong/cruise/mlp-ark/.python-version)
  3.7.4

C02XV41NJHD2:mlp-ark zhanwu.xiong$ which python3.7
/Users/zhanwu.xiong/.pyenv/shims/python3.7

C02XV41NJHD2:mlp-ark zhanwu.xiong$ python3.7
pyenv: python3.7: command not found

The `python3.7' command exists in these Python versions:
  3.7.4
```

Arguably, this could be a `pyenv` issue

Eric-Arellano:
Closes https://github.com/pantsbuild/setup/issues/62.

A frequent idiom with Python development is to use Pyenv to globally install several Python versions, but to only activate certain versions locally. This would break the `./pants` script, as it would think that an un-activated version like `python3.8` was valid when really it would fail when being run.

### DIFF
--- a/pants
+++ b/pants
@@ -116,7 +116,15 @@ function determine_default_python_exe {
     supported_message='3.6+'
   fi
   for version in "${supported_python_versions[@]}"; do
-    command -v "python${version}" && return 0
+    interpreter_path="$(command -v "python${version}")"
+    if [[ -z "${interpreter_path}" ]]; then
+      continue
+    fi
+    # Check if the Python version in installed via Pyenv but not activated.
+    if [[ "$("${interpreter_path}" --version 2>&1 > /dev/null)" == "pyenv: python${version}"* ]]; then
+      continue
+    fi
+    echo "${interpreter_path}" && return 0
   done
   die "No valid Python interpreter found. For this Pants version, Pants requires Python ${supported_message}."
 }
--- a/pants
+++ b/pants
@@ -120,7 +120,7 @@ function determine_default_python_exe {
     if [[ -z "${interpreter_path}" ]]; then
       continue
     fi
-    # Check if the Python version in installed via Pyenv but not activated.
+    # Check if the Python version is installed via Pyenv but not activated.
     if [[ "$("${interpreter_path}" --version 2>&1 > /dev/null)" == "pyenv: python${version}"* ]]; then
       continue
     fi


benjyw (approved):

pants
```diff
@@ -116,7 +116,15 @@ function determine_default_python_exe {
     supported_message='3.6+'
   fi
   for version in "${supported_python_versions[@]}"; do
-    command -v "python${version}" && return 0
+    interpreter_path="$(command -v "python${version}")"
+    if [[ -z "${interpreter_path}" ]]; then
+      continue
+    fi
+    # Check if the Python version is installed via Pyenv but not activated.
```
jsirois, line 123: This was a great step forward - we no longer error.

Better would be to be able ignore shims, but that's hard to do. Best would be to be fully pyenv aware as pants is and skip the shims (via your fix here) and use pyenv pythons even when not activated via a `$PYENV_ROOT/versions/*/bin/python${version}` scan: https://github.com/pantsbuild/pants/blob/c8243f1ca66d00495033783eacfe4c02d76dcfae/src/python/pants/python/python_setup.py#L265-L290

Eric-Arellano: > Better would be to be able ignore shims, but that's hard to do. Best would be to be fully pyenv aware and use pyenv pythons even when not activated

The idea being that it's an implementation detail which Python interpreter is used to run Pants, so it shouldn't matter if a user has disabled the interpreter with `pyenv`? That does not necessarily signal intent that they don't want Pants using that particular interpreter?

There is also always a safety valve via `$PYTHON`.

jsirois:
> so it shouldn't matter if a user has disabled the interpreter with pyenv?

Perhaps, but disable is misleading. Its really the opposite - they have enabled an interpreter (potentially for a particular directory). That implies they want to use a particular interpreter when working on code in a certain directory subtree. Pants though is a tool that runs using the interpreter it wants and this does not violate what the user wants to run against their code, Pants still respects that. So basically, we could do better in the future. For any user that has the python Pants needs to run, but only via pyenv, this enhancement would allow panst to work as they might expect it should, even when they have pyenv setup to use in their current project with an intepreter Pants does not support as its runtime.

PR #67, opened 2020-02-22
Use pyenv to test Python interpreter resolution rather than `--pants-runtime-python-version`

Eric-Arellano:
We previously used `--pants-runtime-python-version` to force Pants to use certain Python versions. This was not safe to keep doing, though, because we ended up removing this option.

Instead, we can use Pyenv to get the same effect.

This will allow us to start testing default Python interpreter resolution for Pants > 1.16.

We remove Ubuntu 12.04 from CI to land this, as it has been EOL since April 2017 and causes issues with getting Pyenv to work properly.

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -33,12 +33,11 @@ class TestBase(TestCase):
                 "`export PYENV_ROOT=$(pyenv root)`."
             )
         self.pyenv_versions = subprocess.run(
-            [self.pyenv_bin, "versions", "--bare"],
+            [self.pyenv_bin, "versions", "--bare", "--skip-aliases"],
             stdout=subprocess.PIPE,
             encoding="utf-8",
             check=True,
         ).stdout.splitlines()
-        print(self.pyenv_versions)
 
     @contextmanager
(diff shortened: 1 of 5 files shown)


tests/test_base.py
```diff
@@ -4,15 +4,41 @@
...
 from contextlib import contextmanager
+from pathlib import Path
 from typing import Iterator, Optional
 from unittest import TestCase
 
 
 class TestBase(TestCase):
     """A base class with useful utils for tests."""
 
+    def setUp(self) -> None:
```
Eric-Arellano, line 18: I tried using `setUpClass` since we only need to do this once-per-class, but for some reason it resulted in a bunch of MyPy issues. MyPy didn't allow us to even do `cls.foo = "example"`..complaining that `foo` is not defined on `Type[TestBase]`.

Because CI is so short (10 minutes), and this setup is quick to execute (<5 seconds), I figured it was better to use `setUp` than have 10 type ignore comments.

benjyw: And we make few changes in this repo anyway, so there aren't many CI runs.

benjyw (approved):

PR #68, opened 2020-02-22
Remove support for Pants <= 1.16.0

Eric-Arellano:
Pants 1.16.0 is over 9 months old. For new Pants users, it is highly unlikely they will change the default version of Pants 1.24.0 all the way back to 1.16.0. 

By removing this support, we can guarantee that Pants is running with Python 3.6+. This allows us to simplify much of the `./pants` code, e.g. allowing us in a followup to use the stdlib `venv` implementation instead of `virtualenv`.

If users try using this script with Pants 1.16.0 or earlier, we point them to a revision that will work properly.

### DIFF
--- a/pants
+++ b/pants
@@ -111,7 +111,6 @@ function determine_default_python_exe {
     fi
     echo "${interpreter_path}" && return 0
   done
-  die "No valid Python interpreter found. Pants requires Python 3.6+ to run."
 }
 
 function determine_python_exe {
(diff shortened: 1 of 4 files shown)


pants
```diff
@@ -73,49 +73,34 @@ EOF
...
-  elif [[ "${pants_major_version}" -eq 1 && "${pants_minor_version}" -eq 17 ]]; then
-    supported_python_versions=('3.6' '3.7' '3.8' '3.9')
-    supported_message='3.6+'
-  else
-    supported_python_versions=('3.9' '3.8' '3.7' '3.6')
-    supported_message='3.6+'
+  if [[ "${pants_major_version}" -eq 1 && "${pants_minor_version}" -le 16 ]]; then
+    die "This version of the \`./pants\` script does not work with Pants <= 1.16.0. Instead,
+either upgrade your \`pants_version\` or use the version of the \`./pants\` script at
+https://raw.githubusercontent.com/pantsbuild/setup/3bb006e4a792ae83d7059ea76c0372ece7c1069d/pants."
```
benjyw, line 97: Thanks for this! Should cover all bases.

benjyw (approved):

PR #69, opened 2020-02-22
Test Python 3.8 in CI

Eric-Arellano:
We claim that Pants works with Python 3.8. We should, therefore, test this.

As part of this change, we remove Ubuntu Trusty and Ubuntu Xenial to only use Ubuntu Bionic, as neither of those come with Python 3.8 by default.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,29 +4,11 @@ branches:
   only:
     - gh-pages
 
-env:
-  global:
-    - PYENV_PY27_VERSION=2.7.17
-    - PYENV_PY36_VERSION=3.6.10
-    - PYENV_PY37_VERSION=3.7.6
-    - PYENV_PY38_VERSION=3.8.1
-    # NB: Linux shards use Pyenv to pre-install Python. We must not override
-    # PYENV_ROOT on those shards, or their Python will no longer work.
-    - PYENV_ROOT="${PYENV_ROOT:-${HOME}/.pants_pyenv}"
-    - PATH="${PYENV_ROOT}/versions/${PYENV_PY27_VERSION}/bin:${PATH}"
-    - PATH="${PYENV_ROOT}/versions/${PYENV_PY36_VERSION}/bin:${PATH}"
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -102,20 +105,12 @@ jobs:
...
-        - SKIP_PYTHON37_TESTS=true
-      before_install:
-        - pyenv global 2.7.14 3.6.3
-
     - name: "Ubuntu 16.04 - Xenial"
       <<: *linux_setup
       dist: xenial
       env:
         - CACHE_NAME="linux.xenial"
+        - SKIP_PYTHON38_TESTS=true
```
jsirois: What's the reason for 2 linux shards? OSX makes sense since they ship historically broken interpreters and there is only 1 shipper, Apple. For Linux though there is a large universe of shippers and Ubuntu is a very small slice. Traditionally we have had ~0 problems with linux distro pythons as well.

A further carrot is Bionic already supports 2.7, 3.6, 3.7 & 3.8 - no need for pyenv shenanigans: https://docs.travis-ci.com/user/reference/bionic/#python-support

jsirois (approved):

.travis.yml
Eric-Arellano: That's fine with me to drop Xenial, but I'm not sure the motivation. The reason I was keeping it is is that there's little cost for us to support it in CI and it slightly improves our coverage.

For example, in the Venv PR, it looks like Trusty succeeded but for some reason Xenial and Bionic both failed (unclear why). It doesn't slow down our CI much and only adds a couple lines of config to have Xenial.

jsirois: The motivation is to drop pyenv and not pretend we cover linux. Its true 2 is more than 1, but its << 100s.

jsirois: > It doesn't slow down our CI much and only adds a couple lines of config to have Xenial.

Power consumption is a real cost we all regularaly ignore since it feels free. it's not. That's probbly the best argument. This is wasteful for no real benefit.

jsirois (approved):

PR #70, opened 2020-02-22
Use std lib `venv` instead of virtualenv

Eric-Arellano:
This results in less code, less logging to users, and better performance.


jsirois (approved):

Eric-Arellano:
Oh. Well, this is a bummer.

```
> eric@eric-VirtualBox:~$ python3 -m venv venv
The virtual environment was not created successfully because ensurepip is not
available.  On Debian/Ubuntu systems, you need to install the python3-venv
package using the following command.

    apt-get install python3-venv

You may need to use sudo with that command.  After installing the python3-venv
package, recreate your virtual environment.

Failing command: ['/home/eric/venv/bin/python3', '-Im', 'ensurepip', '--upgrade', '--default-pip']
```

PR #72, opened 2020-02-23
Convert tests to function style.

jsirois:
This moves tests from `TestCase` subclasses to test functions and pytest
fixtures. Benefits include generally more straight forward typing which
makes it easier to do the right thing for setup/teardown in terms of
fixture scope as well as eliminating most explicit contextmanager use;
thereby reducing nesting.

Pytest has a bit of a magic curve and this change does not document the
magic, but instead expects maintainers to read up a bit and learn about
fixtures dependency injection by name behavior and the pre-defined
fixtures supplied by pytest like `tmp_path` and `monkeypatch`.

This is a follow up to #67:
  https://github.com/pantsbuild/setup/pull/67#discussion_r382932317


jsirois:
This may not be deemed worth it / an improvement. I saw https://github.com/pantsbuild/setup/pull/67#discussion_r382932317 though and I find that sort of thing frustrating so wanted to offer a potential solution in general, and its application here in particular. The magic dependency injection of fixtures is magic and I don't like that, but it does allow things to be arguably more straight forward, that magic aside.

Your call.

tests/conftest.py
```diff
@@ -0,0 +1,42 @@
...
+import shutil
+import subprocess
+from pathlib import Path, PurePath
+from typing import Optional, Protocol
+
+import pytest
+
+
+@pytest.fixture(scope="session")
+def project_root() -> PurePath:
```
jsirois, line 14: This was not technically needed since we run tox from the repo root in practice and I think tox enforces a CWD at tox.ini anyhow. It does always read as magic to me though that relative paths like "pants" work. This just makes that path explicit. May not be worth it / may be just noise.

tests/test_sanity_check.py
```diff
@@ -4,29 +4,94 @@
...
-from test_base import TestBase
+import pytest
+from helpers import create_pants_config
 
 
-class TestSanityCheck(TestBase):
+@pytest.fixture(scope="module")
+def pyenv_bin() -> str:
+    pyenv_bin = os.environ.get("PYENV_BIN") or shutil.which("pyenv")
+    if pyenv_bin is None:
```
Eric-Arellano, line 21: Much, much better.

tests/conftest.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2019 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+import shutil
+import subprocess
+from pathlib import Path, PurePath
+from typing import Optional, Protocol
```
Eric-Arellano: Protocol is only in Python 3.8. I believe you can change `basepython` in `tox.ini` to `python3.8` to fix the CI failures. Otherwise, you'd need to use `typing_extensions`.

Eric-Arellano (approved):
This is an enormous improvement! Thank you!

tests/conftest.py
Eric-Arellano: I prefer the more explicit approach as most people (including me before reading this) don't know this is how Tox works.

tests/test_sanity_check.py
```diff
@@ -35,19 +100,32 @@ def run_command(command: List[str], **kwargs: Any) -> None:
...
+
+def check_for_all_python_versions(
+    sanity_checker: SanityChecker,
+    *python_versions: str,
+    pants_version: Optional[str],
+    use_toml: bool = True,
+) -> None:
+    for python_version in python_versions:
+        sanity_checker.sanity_check(
+            pants_version=pants_version, python_version=python_version, use_toml=use_toml
```
Eric-Arellano: Consider moving this to be a method on `SanityChecker`. Then the test would read:
```python
sanity_checker.sanity_check(python_version=None, pants_version="1.26.0.dev0")
sanity_checker.check_for_all_python_versions("3.6", "3.7", "3.8", pants_version="1.26.0.dev0")
```

tests/conftest.py
jsirois: Right. Went with forcing 3.8.

tests/conftest.py
jsirois: OK, great. Made perfectly explicit now by setting cwd to this file's parent.

tests/test_sanity_check.py
jsirois: Done.

Eric-Arellano (approved):

Issue #74, opened 2020-03-04
Make determining the current pants version independent from system Python


fraprue:
Currently in the function `determine_pants_version` a plain python command is used to fetch the pants version from the JSON blob returned by the curl command.
This obviously won't work if `python` is not available.

In my case I had a Docker image, where Python3.7 was installed and I had to register it as an alternative for the `python` command.

Wouldn't it be better to instead use the same Python executable as returned by `determine_python_exe`?
I changed my pants script locally, so that `determine_pants_version` would take the path to the Python executable as an input argument and called it like this:

    python="$(determine_python_exe)"
    pants_version="$(determine_pants_version "${python}")"

This also seemed to work like a charm.

Eric-Arellano:
Great idea! This was only an oversight why we weren't doing this before. Your tweak is a good one.

Would you have time this week to put up a PR? 

--

You'll need to also reorder the high level flow:

https://github.com/pantsbuild/setup/blob/f963ce9a3d0708fbcf4b83b904725e34135dddd9/pants#L79-L89

I don't think you need to change any of the copy, only re-order it.

PR #75, opened 2020-03-04
Avoid plain python call in determine_pants_version (#74)

fraprue:


Eric-Arellano (approved):
This looks great! Thank you for the fix! 

We'll be sending an email to `pants-devel` today encouraging users to migrate from `pants.ini` to `pants.toml`. To do that, they'll need to re-pull the `./pants` script, which means that they will also end up getting your fix here 🎉

Eric-Arellano:
(Only CI failure is because Travis recently changed the Pyenv versions for Ubuntu Bionic. Going to merge this because every other shard is green. I will fix the issue in a followup PR.)

PR #78, opened 2020-05-24
Fix first_time test: make robust to warnings or other stderr

RemiTorracinta:
Currently, the `test_only_bootstraps_the_first_time` test is failing, as a result of brittleness to warnings and other textual output that is now present when pants runs.

It might be appropriate to update the configuration used when bootstrapping pants so that these warnings and messages are not present, but regardless I think updating the assertions here makes sense and is more future-proof.

Example of build failure:
https://travis-ci.com/github/pantsbuild/setup/jobs/339413866

Sidenote: it's unfortunate that this test broke due to changes in another repo. Might be a good argument for consolidating this setup code into the main pants repository.

### DIFF
--- a/tests/test_first_time_install.py
+++ b/tests/test_first_time_install.py
@@ -39,7 +39,7 @@ def test_only_bootstraps_the_first_time(build_root: Path) -> None:
         encoding="utf-8",
         cwd=str(build_root),
     ).stderr
-    assert first_run_pants_script_logging
+    assert "Collecting pantsbuild.pants==" in first_run_pants_script_logging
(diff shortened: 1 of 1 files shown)


Eric-Arellano (approved):
Thanks for fixing this, Remi! I agree with your assessment that it's better to make the test less brittle.

For what it's worth, we instruct new users to set up a minimal `pants.toml` before ever running `./pants` so that they don't hit the deprecation: https://pants.readme.io/docs/installation#creating-the-launch-script. And the deprecation will end soon.

I'll merge on green.

PR #79, opened 2020-06-12
Parameterize the config file name.

benjyw:
Instead of hardcoding pants.toml/pants.ini.

Allows the script to be shared for both
v2-only and v1-only invocations, by pointing
it at different config.

Also update a test to use pants 1.28.0. That test
no longer passed, even at unmodified HEAD, due to
some watchman issue. In any case, it's more important
to test against a recent pants, since people downloading
the script are likely to be doing so in order to run
the most recent stable release (which is what the setup
script will choose for them on first bootstrap).

### DIFF
--- a/pants
+++ b/pants
@@ -55,6 +55,7 @@ function get_exe_path_or_die {
   fi
 }
 
+pants_config_file=""
 if [[ -f "${PANTS_INI}" ]]; then
   pants_config_file="${PANTS_INI}"
 fi
@@ -204,12 +205,16 @@ pants_dir="$(bootstrap_pants "${pants_version}" "${python}")"
 # See https://blog.phusion.nl/2017/10/13/why-ruby-app-servers-break-on-macos-high-sierra-and-what-can-be-done-about-it/
 export no_proxy='*'
 
-if [[ -z ${pants_config_file+x} ]]; then
+if [[ -z "${pants_config_file}" ]]; then
   pants_config_file_flag="";
 else
   pants_config_file_flag=--pants-config-files="[\"${PWD}/${pants_config_file}\"]";
 fi
 
-set -x
+# MacOS's default open files limit is too low for Pants, bump it to something reasonable.
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  ulimit -n 10000
+fi
+
 exec "${pants_dir}/bin/python" "${pants_dir}/bin/pants" \
   "${pants_config_file_flag}" --pants-bin-name="${PANTS_BIN_NAME}" "$@"
--- a/pants
+++ b/pants
@@ -211,10 +211,5 @@ else
   pants_config_file_flag=--pants-config-files="[\"${PWD}/${pants_config_file}\"]";
 fi
 
-# MacOS's default open files limit is too low for Pants, bump it to something reasonable.
-if [[ "$OSTYPE" == "darwin"* ]]; then
-  ulimit -n 10000
-fi
-
 exec "${pants_dir}/bin/python" "${pants_dir}/bin/pants" \
   "${pants_config_file_flag}" --pants-bin-name="${PANTS_BIN_NAME}" "$@"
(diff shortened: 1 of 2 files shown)


pants
```diff
@@ -51,17 +55,24 @@ function get_exe_path_or_die {
...
+fi
+if [[ -f "${PANTS_TOML}" ]]; then
+  pants_config_file="${PANTS_TOML}"
+fi
+
 function get_pants_config_value {
   config_key="$1"
   optional_space="[[:space:]]*"
-  if [[ -f 'pants.ini' ]]; then
+  if [[ "${pants_config_file}" == *.ini ]]; then
```
stuhood, line 69: I think that this will fail to match if there are multiple ini files? It seems like continuing to match using `-f` makes sense here?

pants
```diff
@@ -15,6 +15,10 @@ set -eou pipefail
 
 PYTHON_BIN_NAME="${PYTHON:-unspecified}"
 
+PANTS_INI=${PANTS_INI:-pants.ini}
+PANTS_TOML=${PANTS_TOML:-pants.toml}
+PANTS_BIN_NAME="${PANTS_BIN_NAME:-$0}"
```
Eric-Arellano, line 20: It would help to explain in a brief comment "Else, default to the name of this script."

pants
```diff
@@ -193,4 +205,16 @@ pants_dir="$(bootstrap_pants "${pants_version}" "${python}")"
...
+if [[ -z "${pants_config_file}" ]]; then
+  pants_config_file_flag="";
+else
+  pants_config_file_flag=--pants-config-files="[\"${PWD}/${pants_config_file}\"]";
+fi
+
+# MacOS's default open files limit is too low for Pants, bump it to something reasonable.
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  ulimit -n 10000
+fi
```
Eric-Arellano: If possible, it'd be better to save this for a followup. This is the type of change that could cause surprising behavior for users so it would help to have a distinct commit for things like `git bisect`.

pants
benjyw: This is a string match. We already checked that this file exists above. This just checks if it has a .ini suffix or a .toml suffix, which is all we care about here.

stuhood: Got it.

stuhood (approved):

pants
Eric-Arellano: Another reason this would be good to do in a followup. If the user already has configured this value to > 10,000, then we should respect what they used.

pants
benjyw: Whether or not there are multiple ini files on disk is not known to this script (wasn't before, and isn't now). We just pick one that we grep the version out of, and this change allows that one to be configurable.

pants
benjyw: I assume the reader knows what `$0` means. I don't like comments that just recapitulate the logic, they are superfluous.

pants
benjyw: Done.

Eric-Arellano (approved):

pants
```diff
@@ -193,4 +205,11 @@ pants_dir="$(bootstrap_pants "${pants_version}" "${python}")"
 # See https://blog.phusion.nl/2017/10/13/why-ruby-app-servers-break-on-macos-high-sierra-and-what-can-be-done-about-it/
 export no_proxy='*'
 
-exec "${pants_dir}/bin/python" "${pants_dir}/bin/pants" "$@"
+if [[ -z "${pants_config_file}" ]]; then
+  pants_config_file_flag="";
+else
+  pants_config_file_flag=--pants-config-files="[\"${PWD}/${pants_config_file}\"]";
```
stuhood, line 211: The effect of this is that if `PANTS_CONFIG_FILES` is set in the environment, this flag will be applied last, leading to something like:
```
pants_config_files = ['/Users/example/repo/pants.toml', '/Users/example/repo/pants.ci.toml', '/Users/example/repo/pants.toml'] (from FLAG from env var PANTS_CONFIG_FILES)
```
... which in this case causes some of the values in `pants.ci.toml` to not to be applied.

PR #80, opened 2020-06-12
Set a higher open file limit on MacOS.

benjyw:
The default is too low for pants usage, and most
users will need to set a higher ulimit, so we might
as well do it for them, and avoid the confusion.

### DIFF
--- a/pants
+++ b/pants
@@ -193,4 +193,9 @@ pants_dir="$(bootstrap_pants "${pants_version}" "${python}")"
 # See https://blog.phusion.nl/2017/10/13/why-ruby-app-servers-break-on-macos-high-sierra-and-what-can-be-done-about-it/
 export no_proxy='*'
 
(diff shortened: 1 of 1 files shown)


stuhood (approved):

Eric-Arellano (commented):
I’m afk so pardon not searching myself - is it possible to see if the user has already configured the value to >10k?

benjyw:
In CI at least we get ```ulimit: open files: cannot modify limit: Operation not permitted```
:(

stuhood:
@benjyw : I keep forgetting, but: I'm pretty sure that we should do this programmatically in pants itself. See https://docs.rs/rlimit/0.3.0/rlimit/enum.Resource.html

benjyw:
> @benjyw : I keep forgetting, but: I'm pretty sure that we should do this programmatically in pants itself. See https://docs.rs/rlimit/0.3.0/rlimit/enum.Resource.html

Sounds good, I will close this for now.

PR #81, opened 2020-06-19
Don't set the config file in the runner script

benjyw:
PANTS_TOML/PANTS_INI are now only used by the script itself, to find the Pants version to bootstrap. If the user also wants Pants to use this as a config file, they must do so explicitly.

### DIFF
--- a/pants
+++ b/pants
@@ -15,8 +15,12 @@ set -eou pipefail
 
 PYTHON_BIN_NAME="${PYTHON:-unspecified}"
 
+# Set these to specify where this script can read the pants version from.
+# This will also be passed to pants itself as its config file, unless the
+# PANTS_CONFIG_FILES var is already set, in which case we defer to that value.
 PANTS_INI=${PANTS_INI:-pants.ini}
 PANTS_TOML=${PANTS_TOML:-pants.toml}
+
 PANTS_BIN_NAME="${PANTS_BIN_NAME:-$0}"
 
 PANTS_HOME="${PANTS_HOME:-${XDG_CACHE_HOME:-$HOME/.cache}/pants/setup}"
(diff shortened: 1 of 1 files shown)


pants
```diff
@@ -205,11 +209,8 @@ pants_dir="$(bootstrap_pants "${pants_version}" "${python}")"
 # See https://blog.phusion.nl/2017/10/13/why-ruby-app-servers-break-on-macos-high-sierra-and-what-can-be-done-about-it/
 export no_proxy='*'
 
-if [[ -z "${pants_config_file}" ]]; then
-  pants_config_file_flag="";
-else
-  pants_config_file_flag=--pants-config-files="[\"${PWD}/${pants_config_file}\"]";
+if ! [[ -v PANTS_CONFIG_FILES ]] && [[ -n ${pants_config_file} ]]; then
+  export PANTS_CONFIG_FILES="[\"${PWD}/${pants_config_file}\"]";
```
Eric-Arellano: Should this be a plus? What happens if a user appends via the CLI?

Eric-Arellano (approved):
Thanks

pants
benjyw: The CLI value will be appended to this.

We want to override the default here, so I don't think we want a +.

pants
```diff
@@ -205,11 +209,8 @@ pants_dir="$(bootstrap_pants "${pants_version}" "${python}")"
 # See https://blog.phusion.nl/2017/10/13/why-ruby-app-servers-break-on-macos-high-sierra-and-what-can-be-done-about-it/
 export no_proxy='*'
 
-if [[ -z "${pants_config_file}" ]]; then
-  pants_config_file_flag="";
-else
-  pants_config_file_flag=--pants-config-files="[\"${PWD}/${pants_config_file}\"]";
+if ! [[ -v PANTS_CONFIG_FILES ]] && [[ -n ${pants_config_file} ]]; then
```
stuhood: If someone sets `PANTS_CONFIG_FILES` to an append value, this won't do the right thing... but I don't see a great way to do this without parsing the config file. Maybe not the end of the world.

Is it worth adding something native to handle this case? I don't know exactly what it would look like...

stuhood (approved):

asherf (approved):

pants
benjyw: I mean, TBH we can also just not do this at all, and say "If you set PANTS_TOML or PANTS_INI that only tells the runner script where to look for the pants version, you still have to tell pants itself to use that file some other way". Maybe that's better/simpler? If you're manipulating PANTS_TOML/PANTS_INI you're already doing something advanced.

benjyw: This was supposed to be a convenience, but it's turning into an inconvenience...

benjyw: Yeah, I think I've convinced myself.

Eric-Arellano: Yeah, I agree with you.

Eric-Arellano (approved):
Thanks. PR title and description need updates.

pants
benjyw: OK, simplified this. PTAL.

Eric-Arellano (approved):

stuhood (approved):
Perfectly reasonable. Thanks!

PR #83, opened 2020-07-17
Allow $PANTS_HOME to be a relative path

Eric-Arellano:
A user requested this because their CI provider GitLab works better with relative paths than absolute paths. Generally, Pants allows relative paths, so it's consistent for us to support them here, too.

This also makes some other improvements:
* Fix our version check to look for >= 1.23, rather than > 1.16. This has been the case, but we didn't change the version check.
* Use the more inclusive term `smoke screen`.

### DIFF
--- a/pants
+++ b/pants
@@ -118,11 +118,6 @@ function determine_pants_version {
   fi
   pants_major_version="$(echo "${pants_version}" | cut -d '.' -f1)"
   pants_minor_version="$(echo "${pants_version}" | cut -d '.' -f2)"
-  if [[ "${pants_major_version}" -eq 1 && "${pants_minor_version}" -le 16 ]]; then
-    die "This version of the \`./pants\` script does not work with Pants <= 1.16.0. Instead,
-either upgrade your \`pants_version\` or use the version of the \`./pants\` script at
-https://raw.githubusercontent.com/pantsbuild/setup/3bb006e4a792ae83d7059ea76c0372ece7c1069d/pants."
-  fi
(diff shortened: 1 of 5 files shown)


pants
```diff
@@ -116,8 +121,8 @@ function determine_pants_version {
   fi
   pants_major_version="$(echo "${pants_version}" | cut -d '.' -f1)"
   pants_minor_version="$(echo "${pants_version}" | cut -d '.' -f2)"
-  if [[ "${pants_major_version}" -eq 1 && "${pants_minor_version}" -le 16 ]]; then
-    die "This version of the \`./pants\` script does not work with Pants <= 1.16.0. Instead,
+  if [[ "${pants_major_version}" -eq 1 && "${pants_minor_version}" -lt 23 ]]; then
+    die "This version of the \`./pants\` script does not work with Pants < 1.23.0. Instead,
 either upgrade your \`pants_version\` or use the version of the \`./pants\` script at
 https://raw.githubusercontent.com/pantsbuild/setup/3bb006e4a792ae83d7059ea76c0372ece7c1069d/pants."
```
stuhood: Should this now recommend a "more recent old version"?

stuhood (approved):
Thanks!

pants
Eric-Arellano: I'm not sure how to handle it. The most robust approach would be to have two checks. First check if < 16, and stick with this message. Then, check if >=16 and < 23, then do a new message.

Thoughts?

stuhood: It would probably be ok to walk a linked list of "try this script, it will recommend an older one, rinse repeat".

PR #84, opened 2020-08-04
Support running from an arbitrary master sha

benjyw:


Eric-Arellano (approved):
Excellent!

pants
```diff
@@ -180,8 +197,14 @@ function bootstrap_venv {
...
+  pants_sha="${3:-}"
 
   pants_requirement="pantsbuild.pants==${pants_version}"
+  if [ -z "${pants_sha}" ]; then
+    maybe_find_links=""
+  else
+    maybe_find_links=" --find-links=https://binaries.pantsbuild.org/wheels/pantsbuild.pants/${pants_sha}/${pants_version/+/%2B}/index.html"
+   fi
   python_major_minor_version="$(get_python_major_minor_version "${python}")"
   target_folder_name="${pants_version}_py${python_major_minor_version}"
```
stuhood, line 217: There is one other change necessary here to allow `pants_requirements` targets to resolve correctly, which is adding the findlinks entry as a `--python-repos-repos` entry: see recent changes to our copy of the script.

benjyw: Good catch, done.

stuhood (approved):
Thanks!

benjyw:
Added a test to prove the pants_requirement() thing.

Eric-Arellano (approved):

PR #85, opened 2020-08-19
Compute version from PANTS_SHA.

benjyw:
Instead of requiring it to be edited in pants.toml, when
running from a custom SHA.

Also ensures that bootstrapping errors exit eagerly, so we
don't create a malformed bootstrap dir.

### DIFF
--- a/pants
+++ b/pants
@@ -21,19 +21,8 @@ PYTHON_BIN_NAME="${PYTHON:-unspecified}"
 PANTS_INI=${PANTS_INI:-pants.ini}
 PANTS_TOML=${PANTS_TOML:-pants.toml}
 
-# NOTE: To use an unreleased version of Pants from the pantsbuild/pants master branch, you need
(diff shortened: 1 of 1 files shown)


pants
```diff
@@ -126,6 +115,12 @@ EOF
 # are installed and up to date.
 
 function determine_pants_version {
+
+  if [ -n "${PANTS_SHA:-}" ]; then
+    get_version_for_sha "$PANTS_SHA"
```
tdyas, line 121: Why retrieve the version if it is not stored (since it won't be used anywhere)?

stuhood: This function "returns" it, by virtue of echoing it. Yay bash.

stuhood (approved):
Thank you!

pants
tdyas: Doh, and I am the one who wrote this function in the version in the toolchain repo ...

@benjyw: Probably worth a comment since this usage confused even me the original author ...

benjyw: Done.

PR #86, opened 2020-09-24
Require `pants_version` to be set

Eric-Arellano:
We generally highly encourage users to use lockfiles because it's important for deterministic builds. But with the Pants script itself, we did not require this determinism, which could result in builds breaking randomly when we do new stable releases.

Our docs already have been saying to set the `pants_version` before running `./pants`, so we don't need to update anything there.

### DIFF
--- a/pants
+++ b/pants
@@ -105,9 +105,8 @@ EOF
 # The high-level flow:
 # 1.) Resolve the Python interpreter, first reading from the env var $PYTHON,
 #     then defaulting to Python 3.6+.
-# 2.) Resolve the Pants version from config or from the latest stable
-#     release to PyPI, so that we know what to name the venv (virtual environment) folder and what
-#     to install with Pip.
+# 2.) Resolve the Pants version from config so that we know what to name the venv
+#     (virtual environment) folder and what to install with Pip.
 # 3.) Check if the venv already exists via a naming convention, and create the venv if not found.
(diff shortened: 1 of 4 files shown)


pants
```diff
@@ -126,8 +125,9 @@ function determine_pants_version {
 
   pants_version="$(get_pants_config_value 'pants_version')"
   if [[ -z "${pants_version}" ]]; then
-    pants_version="$(curl -sSL https://pypi.python.org/pypi/pantsbuild.pants/json |
-      "${python}" -c "import json, sys; print(json.load(sys.stdin)['info']['version'])")"
+    die "Please explicitly specify the \`pants_version\` in your \`pants.toml\` under the
```
stuhood, line 128: Could we warn with the version that we used instead?

Eric-Arellano: That warning would happen every single run, which is super noisy. I don't think that would be that much better of a UX. What's your thinking for not wanting to fail fast?

tdyas: How does this interact with PANTS_SHA?

Eric-Arellano: Right above, we have:

```
  if [ -n "${PANTS_SHA:-}" ]; then
    # get_version_for_sha will echo the version, thus "returning" it from this function.
    get_version_for_sha "$PANTS_SHA"
    return
  fi
```

So it's good to go.

stuhood (approved):

PR #88, opened 2020-11-11
Remove support for pants.ini config file

Eric-Arellano:
Pants 2.x no longer supports INI.

PR #89, opened 2020-11-12
Hotfix CI failure from removing pants.ini support

Eric-Arellano:
For some reason, CI didn't trigger on https://github.com/pantsbuild/setup/pull/88.

This also updates our check for using outdated Pants versions to require the first version to support `pants.toml`: 1.26.


tdyas (approved):
lgtm

PR #90, opened 2020-11-12
Pants 2.2 will no longer run with Python 3.6

Eric-Arellano:
We now choose and validate the Python interpreter based on the Pants version.

* Pants 1.x and 2.0 will continue to prefer 3.6, then 3.7, then 3.8, as 3.6 is the best tested.
    * It would be better to prefer 3.8 because it means Pants will work on Py38 code with dep inference, but there are some known Pantsd issues with Py38.
* Pants 2.1 will prefer 3.7, then 3.8, then the deprecated 3.6.
* Pants 2.2+ will require 3.7 or 3.8.

We will likely add 3.9 support soon, but we're not sure which version. Rather than speculating, we will update this script when the time is right.

### DIFF
--- a/pants
+++ b/pants
@@ -6,7 +6,6 @@
 # This ./pants bootstrap script comes from the pantsbuild/setup
 # project. It is intended to be checked into your code repository so
 # that other developers have the same setup.
-
 #
 # Learn more here: https://www.pantsbuild.org/docs/installation
 # ====================================================================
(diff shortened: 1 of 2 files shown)


pants
```diff
@@ -62,23 +61,24 @@ function tempdir {
 }
 
 function get_exe_path_or_die {
-  exe="$1"
+  local exe="$1"
```
Eric-Arellano, line 64: I didn't realize that function variables are global by default...I added `local` to lots of places, pardon the noise.

pants
```diff
@@ -124,8 +122,34 @@ at https://raw.githubusercontent.com/Eric-Arellano/setup/0d445edef57cb89fd830db7
   echo "${pants_version}"
 }
 
+function set_supported_python_versions {
```
Eric-Arellano, line 125: Returning `Tuple[List[str], List[str], str]` was far too complex in Bash, so mutation is the best I could come up with.

pants
```diff
@@ -124,8 +122,34 @@ at https://raw.githubusercontent.com/Eric-Arellano/setup/0d445edef57cb89fd830db7
...
+  local pants_version="$1"
+  local pants_major_version
+  local pants_minor_version
+  pants_major_version="$(echo "${pants_version}" | cut -d '.' -f1)"
+  pants_minor_version="$(echo "${pants_version}" | cut -d '.' -f2)"
```
Eric-Arellano, line 130: I tested that this does still work properly with PANTS_SHA :)

tdyas (approved):
lgtm

jsirois (approved):
> It would be better to prefer 3.8 because it means Pants will work on Py38 code with dep inference, but there are some known Pantsd issues with Py38

That keeps coming up and it keeps pointing to not cheating being the only reasonable path forward. Getting inference to work for python via Process execution seems critical.

Eric-Arellano:
> Getting inference to work for python via Process execution seems critical.

Yeah, likely. We could use `interpreter_constraints` to know which interpreter to use for that particular file, rather than the current approach of trying multiple ASTs.

Otherwise, we are in a never-ending race to have Pants run on the latest Python release for dep inference to work properly. For example, we won't be able to parse 3.10-dev because it makes syntax changes.

PR #95, opened 2021-01-29
removed no_proxy from pants launch script

cristianmatache:
Setting no_proxy in the pants script overwrites the user-defined value of the variable. I am not sure if this needed anymore.
If it is we can change it to an if statement or something like that.


Eric-Arellano (approved):
Indeed, we don't use the `requests` library anymore. Pretty sure this is stale. Thanks!

PR #97, opened 2021-05-24
Use latest version of virtualenv

benjyw:
For compatibility with Python3.9.  We now consume virtualenv as a zipapp, as the previous method no longer works.

We also now check virtualenv's sha256, for added security. 
This change also fixes some cases where we weren't exiting an outer process when a subprocess exited on error.

### DIFF
--- a/pants
+++ b/pants
@@ -40,6 +40,7 @@ PANTS_BOOTSTRAP="${PANTS_SETUP_CACHE}/bootstrap-$(uname -s)-$(uname -m)"
 # See https://virtualenv.pypa.io/en/latest/installation.html#via-zipapp
 VIRTUALENV_ZIPAPP=virtualenv.pyz
 VIRTUALENV_ZIPAPP_URL="https://bootstrap.pypa.io/virtualenv/3.6/${VIRTUALENV_ZIPAPP}"
+VIRTUALENV_ZIPAPP_EXPECTED_SHA256="9d5c8b09cdeaf44414f07902d2ae45c4bb8a2085717d66f82ae389a9c6f94425"
 
 COLOR_RED="\x1b[31m"
 COLOR_GREEN="\x1b[32m"
@@ -110,9 +111,9 @@ function determine_pants_version {
 
   pants_version="$(get_pants_config_value 'pants_version')"
   if [[ -z "${pants_version}" ]]; then
-    die "Please explicitly specify the \`pants_version\` in your \`pants.toml\` under the
-\`[GLOBAL]\` scope. See https://pypi.org/project/pantsbuild.pants/#history for all released
-versions and https://www.pantsbuild.org/docs/installation for more instructions."
+    die "Please explicitly specify the \`pants_version\` in your \`pants.toml\` under the \`[GLOBAL]\` scope.
+See https://pypi.org/project/pantsbuild.pants/#history for all released versions
+and https://www.pantsbuild.org/docs/installation for more instructions."
   fi
   pants_major_version="$(echo "${pants_version}" | cut -d '.' -f1)"
   pants_minor_version="$(echo "${pants_version}" | cut -d '.' -f2)"
@@ -186,7 +187,7 @@ function determine_python_exe {
     fi
   fi
   local python_exe
-  python_exe="$(get_exe_path_or_die "${python_bin_name}")"
+  python_exe="$(get_exe_path_or_die "${python_bin_name}")" || exit 1
   local major_minor_version
   major_minor_version="$(get_python_major_minor_version "${python_exe}")"
   for valid_version in "${supported_python_versions_int[@]}"; do
@@ -209,9 +210,14 @@ function bootstrap_virtualenv {
       staging_dir=$(tempdir "${PANTS_BOOTSTRAP}")
       cd "${staging_dir}"
       curl -LO "${VIRTUALENV_ZIPAPP_URL}"
+      fingerprint="$(shasum -a 256 "${VIRTUALENV_ZIPAPP}" | cut -d' ' -f1)"
+      if [[ "${VIRTUALENV_ZIPAPP_EXPECTED_SHA256}" != "${fingerprint}" ]]; then
+        die "SHA256 of ${staging_dir}/${VIRTUALENV_ZIPAPP} is not as expected. Aborting."
+      fi
       ln -s "${staging_dir}/${VIRTUALENV_ZIPAPP}" "${staging_dir}/latest"
       mv "${staging_dir}/latest" "${PANTS_BOOTSTRAP}/${VIRTUALENV_ZIPAPP}"
-    ) 1>&2
+      green "SHA256 fingerprint of "${PANTS_BOOTSTRAP}/${VIRTUALENV_ZIPAPP}" verified."
+    ) 1>&2 || exit 1
   fi
   echo "${PANTS_BOOTSTRAP}/${VIRTUALENV_ZIPAPP}"
 }
@@ -265,7 +271,7 @@ function bootstrap_pants {
         # The venv module is not available, so download and use the virtualenv zipapp.
         green "${python} does not have the venv module avilable, using virtualenv instead."
         local virtualenv_path
-        virtualenv_path="$(bootstrap_virtualenv)"
+        virtualenv_path="$(bootstrap_virtualenv)" || exit 1
         virtualenv_cmd="${python} ${virtualenv_path} --no-download"
       fi
 
@@ -277,7 +283,7 @@ function bootstrap_pants {
       ln -s "${staging_dir}/install" "${staging_dir}/${target_folder_name}" && \
       mv "${staging_dir}/${target_folder_name}" "${PANTS_BOOTSTRAP}/${target_folder_name}" && \
       green "New virtual environment successfully created at ${PANTS_BOOTSTRAP}/${target_folder_name}."
-    ) 1>&2
+    ) 1>&2 || exit 1
   fi
   echo "${PANTS_BOOTSTRAP}/${target_folder_name}"
 }
@@ -286,7 +292,8 @@ function bootstrap_pants {
 cd "$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
 pants_version="$(determine_pants_version)"
 python="$(determine_python_exe "${pants_version}")"
-pants_dir="$(bootstrap_pants "${pants_version}" "${python}" "${PANTS_SHA:-}")"
+pants_dir="$(bootstrap_pants "${pants_version}" "${python}" "${PANTS_SHA:-}")" || exit 1
+
 pants_python="${pants_dir}/bin/python"
 pants_binary="${pants_dir}/bin/pants"
 pants_extra_args=""
--- a/pants
+++ b/pants
@@ -35,7 +35,8 @@ fi
 PANTS_BOOTSTRAP="${PANTS_SETUP_CACHE}/bootstrap-$(uname -s)-$(uname -m)"
 
 VIRTUALENV_ZIPAPP=virtualenv.pyz
-VIRTUALENV_ZIPAPP_URL="https://raw.githubusercontent.com/pypa/get-virtualenv/20.4.7/public/${VIRTUALENV_ZIPAPP}"
+VIRTUALENV_VERSION=20.4.7
+VIRTUALENV_ZIPAPP_URL="https://raw.githubusercontent.com/pypa/get-virtualenv/${VIRTUALENV_VERSION}/public/${VIRTUALENV_ZIPAPP}"
 VIRTUALENV_ZIPAPP_EXPECTED_SHA256="9d5c8b09cdeaf44414f07902d2ae45c4bb8a2085717d66f82ae389a9c6f94425"
 
 COLOR_RED="\x1b[31m"
@@ -214,8 +215,8 @@ EOF
 
 function bootstrap_virtualenv {
   local python="$1"
-  local bootstrapped="${PANTS_BOOTSTRAP}/${VIRTUALENV_ZIPAPP}"
-  if [[ ! -d "${bootstrapped}" ]]; then
+  local bootstrapped="${PANTS_BOOTSTRAP}/virtualenv-${VIRTUALENV_VERSION}/${VIRTUALENV_ZIPAPP}"
+  if [[ ! -f "${bootstrapped}" ]]; then
     (
       green "Downloading the virtualenv zipapp"
       mkdir -p "${PANTS_BOOTSTRAP}"
@@ -228,6 +229,7 @@ function bootstrap_virtualenv {
         die "SHA256 of ${VIRTUALENV_ZIPAPP_URL} is not as expected. Aborting."
       fi
       green "SHA256 fingerprint of ${VIRTUALENV_ZIPAPP_URL} verified."
+      mkdir -p "$(dirname "${bootstrapped}")"
       mv -f "${staging_dir}/${VIRTUALENV_ZIPAPP}" "${bootstrapped}"
       rmdir "${staging_dir}"
     ) 1>&2 || exit 1
--- a/pants
+++ b/pants
@@ -213,8 +213,8 @@ EOF
 # functions.  Any tmp dir w/o a symlink pointing to it can go.
 
 function bootstrap_virtualenv {
-  python="$1"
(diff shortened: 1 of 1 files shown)


benjyw:
@jsirois checking that this is OK, in case we bootstrapped the full `virtualenv` tool for a reason.

Eric-Arellano (approved):
I tried doing this in https://github.com/pantsbuild/setup/pull/70 and I gave up because `venv` wasn't pre-installed in some environments. But imo, this is a good change to make and we should have landed #70 last year.

This has an extra benefit: one fewer download to make, improving our supply chain security.

Eric-Arellano:
> in case we bootstrapped the full virtualenv tool for a reason.

Mostly legacy, other than issue in #70. Python 2 did not have builtin venv.

jsirois:
Yeah, this will need to be more sophisticated. Beyond Python 2 needing virtualenv we have the fairly common problem of Linux distributions that do not include the venv module by default. Debian distros do this and this is similar to https://github.com/pantsbuild/pex/issues/1027. I think we either need to use a more modern virtualenv or else try `-mvenv` and either fall back to virtualenv or else, similar to Pex, link an issue that explains why the venv module might be missing / how to get it in some known cases.

Eric-Arellano:
> Beyond Python 2 needing virtualenv

To be clear, that's a non-issue for Pants. This script has long assumed you're running Pants with Py3.

jsirois:
Ack. But also to be clear, the missing venv module under Python 3 is real. As such we were correct to never land #70.

benjyw:
Hrmph, OK, will think on this.

benjyw:
So, upgrading virtualenv leads to other complications: the latest versions no longer have the root `virtualenv.py` script. They now strongly expect you to install via pip or similar. Which leads to a whole other can of worms, since we don't know if we're even able to do so. 

So given this, plus the fact that we want the setup script to be as simple as possible, one option is to require venv to be available in the python interpreter, check for this, and give a useful error message. We already require an interpreter, so requiring one with minimal functionality doesn't seem awful. Thoughts?

Longer term, packaging Pants with its interpreter (a la PyOxidizer or similar) seems like the way to go.

jsirois:
Works for me I guess. As long as we acknowledge we're solving one user's issue and opening up a new missing venv issue for pre-existing users on their next Pants version bump in CI (I assume the majority of Debian use cases are CI).

benjyw:
> Works for me I guess. As long as we acknowledge we're solving one user's issue and opening up a new missing venv issue for pre-existing users on their next Pants version bump in CI (I assume the majority of Debian use cases are CI).

Well, only if they upgrade the script, right? Just upgrading Pants with the existing runner script should be OK (until they try to upgrade to Python 3.9).

At present we're seeing at least two new users hit issues with this current script and Python 3.9, and using a newer virtualenv seems tricky, unless we're able to pip-install it. Which we could make a best-effort attempt to do I guess?

Eric-Arellano:
> Well, only if they upgrade the script, right?

Right

jsirois:
Correct, but they need to upgrade the script if they have users bitten by the bug you're fixing here. There is a push-pull. There is 1 pants script per repo. That repo will be used by desktop users and CI. It would be awesome not to force a repo to pick one of these to not break and I stead not break either.

benjyw:
Fair point. So should we try to directly pip-install virtualenv? What if `pip` isn't present? It does seem like recent versions of virtualenv cannot be downloaded and run directly, as we do today.

jsirois:
Well, simply using the status quo virtualenv as a fallback would strictly move us forward. We'd cover all the old cases as well as the motivating issue here.

That said, did you check out the virtualenv installation docs? https://virtualenv.pypa.io/en/latest/installation.html#via-zipapp

benjyw:
I did not read them in enough detail! I didn't know what zipapp was but assumed it was some tool we would need to bootstrap in turn. This is promising.

benjyw:
OK, made a change that I think combines the best of all worlds. LMKWYT.

pants
```diff
@@ -34,10 +34,12 @@ fi
...
-VENV_VERSION=${VENV_VERSION:-16.4.3}
-
-VENV_PACKAGE=virtualenv-${VENV_VERSION}
-VENV_TARBALL=${VENV_PACKAGE}.tar.gz
+# Note that this URL points to the zipapp for the latest version of virtualenv
+# that supports the Python version specified in the URL. So it can be set to
+# the lowest version of Python we support running Pants on.
+# See https://virtualenv.pypa.io/en/latest/installation.html#via-zipapp
+VIRTUALENV_ZIPAPP=virtualenv.pyz
+VIRTUALENV_ZIPAPP_URL="https://bootstrap.pypa.io/virtualenv/3.6/${VIRTUALENV_ZIPAPP}"
```
jsirois: The lowest version we currently support, IIUC, is 3.7. Is 3.6 here to support older versions of Pants? If so can you note how old?

pants
```diff
@@ -250,12 +253,25 @@ function bootstrap_pants {
 
   if [[ ! -d "${PANTS_BOOTSTRAP}/${target_folder_name}" ]]; then
     (
-      local venv_path
-      venv_path="$(bootstrap_venv)"
+      green "Bootstrapping Pants using ${python}"
       local staging_dir
       staging_dir=$(tempdir "${PANTS_BOOTSTRAP}")
+
+      if ${python} -c "import venv" > /dev/null  2>&1; then
```
jsirois: I think you want to quote `${python}` - people do strange things.

pants
```diff
@@ -250,12 +253,25 @@ function bootstrap_pants {
...
+        # It is safer and faster to use venv, so we do so when possible.
+        green "${python} has the venv module available, using it."
+        virtualenv_cmd="${python} -m venv"
+      else
+        # The venv module is not available, so download and use the virtualenv zipapp.
+        green "${python} does not have the venv module avilable, using virtualenv instead."
```
jsirois: Spelling on available.

jsirois (approved):

pants
benjyw: Yep, see https://github.com/pantsbuild/setup/blob/gh-pages/pants#L125  which has logic to map pants versions to supported python versions. 

But note that the current virtualenv release still supports all the way back to 2.7, so we're good for a while...

pants
benjyw: Done.

pants
benjyw: Fixed.

benjyw:
Addressed the comments, and also added SHA256 checking for the virtualenv zipapp. Might as well.

pants
```diff
@@ -197,20 +201,25 @@ function determine_python_exe {
...
       local staging_dir
       staging_dir=$(tempdir "${PANTS_BOOTSTRAP}")
       cd "${staging_dir}"
-      curl -LO "https://pypi.io/packages/source/v/virtualenv/${VENV_TARBALL}"
-      tar -xzf "${VENV_TARBALL}"
-      ln -s "${staging_dir}/${VENV_PACKAGE}" "${staging_dir}/latest"
-      mv "${staging_dir}/latest" "${PANTS_BOOTSTRAP}/${VENV_PACKAGE}"
-    ) 1>&2
+      curl -LO "${VIRTUALENV_ZIPAPP_URL}"
+      fingerprint="$(shasum -a 256 "${VIRTUALENV_ZIPAPP}" | cut -d' ' -f1)"
```
jsirois: Hrm. Do we have this dependency already? We could just use python for this since we already depend on it.

benjyw: Yeah good point. Might as well.

benjyw: Done.

jsirois (approved):

pants
```diff
@@ -34,10 +34,13 @@ fi
 
 PANTS_BOOTSTRAP="${PANTS_SETUP_CACHE}/bootstrap-$(uname -s)-$(uname -m)"
 
-VENV_VERSION=${VENV_VERSION:-16.4.3}
-
-VENV_PACKAGE=virtualenv-${VENV_VERSION}
-VENV_TARBALL=${VENV_PACKAGE}.tar.gz
+# Note that this URL points to the zipapp for the latest version of virtualenv
```
Eric-Arellano: > for the latest version of virtualenv

Is the SHA going to be stable then? If a new release comes out, I'd expect that to change and break people's builds.

pants
```diff
@@ -194,23 +198,44 @@ function determine_python_exe {
...
       cd "${staging_dir}"
-      curl -LO "https://pypi.io/packages/source/v/virtualenv/${VENV_TARBALL}"
-      tar -xzf "${VENV_TARBALL}"
-      ln -s "${staging_dir}/${VENV_PACKAGE}" "${staging_dir}/latest"
-      mv "${staging_dir}/latest" "${PANTS_BOOTSTRAP}/${VENV_PACKAGE}"
-    ) 1>&2
+      curl -LO "${VIRTUALENV_ZIPAPP_URL}"
+      fingerprint="$(compute_sha256 "${python}" "${VIRTUALENV_ZIPAPP}")"
+      if [[ "${VIRTUALENV_ZIPAPP_EXPECTED_SHA256}" != "${fingerprint}" ]]; then
+        die "SHA256 of ${staging_dir}/${VIRTUALENV_ZIPAPP} is not as expected. Aborting."
```
Eric-Arellano: Maybe explicitly reference `VIRTUALENV_ZIPAPP_URL`? As a user, I'm not sure I would understand how to fix this issue.

Eric-Arellano (commented):
Thanks!

pants
benjyw: Oh you make an excellent point. It will break on the next virtualenv upgrade. Damn. Good catch!

benjyw: Well, crap. There is no way that I can see to download a zipapp at a specific version. It's like they actively don't want you to verify these.

benjyw: Oh, there is: ```If you are looking for past version of virtualenv.pyz they are available here: https://github.com/pypa/get-virtualenv/blob/<virtualenv version>/public/<python version>/virtualenv.pyz?raw=true```

benjyw: Phew

Eric-Arellano: > If you are looking for past version of virtualenv.pyz they are available here: https://github.com/pypa/get-virtualenv/blob/<virtualenv version>/public/<python version>/virtualenv.pyz?raw=true

benjyw:
OK a few more changes:

- Get a fixed version of virtualenv.
- Don't even bother trying to use venv if available, as it turns out (see Travis failures) that it doesn't always install pip into the virtual environment, even though it is supposed to. virtualenv doesn't appear to have this problem (as I tested by sshing into the Travis container and trying things out). 
- A few improvements of var assignments and stuff.

Eric-Arellano (approved):
Excellent, thanks Benjy

pants
```diff
@@ -245,32 +267,35 @@ function bootstrap_pants {
...
-      venv_path="$(bootstrap_venv)"
+      green "Bootstrapping Pants using ${python}"
       local staging_dir
       staging_dir=$(tempdir "${PANTS_BOOTSTRAP}")
+      local virtualenv_path
+      virtualenv_path="$(bootstrap_virtualenv "${python}")" || exit 1
+      green "Installing ${pants_requirement} into a virtual environment at ${bootstrapped}"
       # shellcheck disable=SC2086
-      "${python}" "${venv_path}/virtualenv.py" --no-download "${staging_dir}/install" && \
+      ${python} "${virtualenv_path}" --no-download "${staging_dir}/install" && \
```
jsirois: The quoting of `${python}` was correct. `/Users/Jane Doe/...` as a home for python is both valid and likelyish.

jsirois (approved):

benjyw:
@jsirois do these errors ring any bells? https://travis-ci.com/github/pantsbuild/setup/jobs/508731786

jsirois:
Nope. But I fiddled. Its not pyenv shims which was my 1st suspect. It is the symlink to the zippapp. I can repro only when I set up one of those and use it:
```
~/Downloads $ mkdir -p some/other/dir
~/Downloads $ ln -s $PWD/virtualenv.pyz some/other/dir/
~/Downloads $ ls -l some/other/dir/
total 0
lrwxrwxrwx 1 jsirois jsirois 38 May 27 13:43 virtualenv.pyz -> /home/jsirois/Downloads/virtualenv.pyz
~/Downloads $ python some/other/dir/virtualenv.pyz bob
Traceback (most recent call last):
  File "/home/jsirois/.pyenv/versions/3.8.1/lib/python3.8/runpy.py", line 193, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/jsirois/.pyenv/versions/3.8.1/lib/python3.8/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "some/other/dir/virtualenv.pyz/__main__.py", line 168, in <module>
  File "some/other/dir/virtualenv.pyz/__main__.py", line 164, in run
  File "some/other/dir/virtualenv.pyz/virtualenv/__main__.py", line 18, in run
  File "some/other/dir/virtualenv.pyz/virtualenv/run/__init__.py", line 32, in cli_run
  File "some/other/dir/virtualenv.pyz/virtualenv/run/session.py", line 46, in run
  File "some/other/dir/virtualenv.pyz/virtualenv/run/session.py", line 53, in _create
  File "some/other/dir/virtualenv.pyz/virtualenv/create/creator.py", line 171, in run
  File "some/other/dir/virtualenv.pyz/virtualenv/create/via_global_ref/builtin/via_global_self_do.py", line 101, in create
  File "some/other/dir/virtualenv.pyz/virtualenv/create/via_global_ref/api.py", line 89, in create
  File "some/other/dir/virtualenv.pyz/virtualenv/create/via_global_ref/api.py", line 92, in install_patch
  File "some/other/dir/virtualenv.pyz/virtualenv/create/via_global_ref/builtin/cpython/cpython3.py", line 27, in env_patch_text
  File "some/other/dir/virtualenv.pyz/virtualenv/create/via_global_ref/api.py", line 103, in env_patch_text
  File "/home/jsirois/.pyenv/versions/3.8.1/lib/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "some/other/dir/virtualenv.pyz/virtualenv/app_data/base.py", line 58, in ensure_extracted
  File "/home/jsirois/.pyenv/versions/3.8.1/lib/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "some/other/dir/virtualenv.pyz/virtualenv/app_data/via_disk_folder.py", line 83, in extract
  File "some/other/dir/virtualenv.pyz/virtualenv/util/zipapp.py", line 22, in extract
    
  File "/home/jsirois/.pyenv/versions/3.8.1/lib/python3.8/zipfile.py", line 1439, in getinfo
    raise KeyError(
KeyError: "There is no item named 'virtualenv.pyz/virtualenv/create/via_global_ref/_virtualenv.py' in the archive"
```

benjyw:
OK, one more change to get rid of the symlink. It seems unnecessary anyway, and I think I just cargo-culted it.

pants
```diff
@@ -228,8 +228,8 @@ function bootstrap_virtualenv {
         die "SHA256 of ${VIRTUALENV_ZIPAPP_URL} is not as expected. Aborting."
       fi
       green "SHA256 fingerprint of ${VIRTUALENV_ZIPAPP_URL} verified."
-      ln -s "${staging_dir}/${VIRTUALENV_ZIPAPP}" "${staging_dir}/latest"
-      mv "${staging_dir}/latest" "${bootstrapped}"
+      mv -f "${staging_dir}/${VIRTUALENV_ZIPAPP}" "${bootstrapped}"
```
jsirois, line 233: I just noticed this but `VIRTUALENV_ZIPAPP` is just the unqualified `virtualenv.pyz`; so bumping the version we download / changing the expected sha256 going forward will no-op above on line 218. So that needs a fix. At that point it may make sense to house the downloaded `virtualenv.pyz` in a directory with its version or hash or both as part of the naming scheme. Then here make a hard link instead of doing a mv.
This way multiple clones of different pants-using repos won't thrash the cache.

benjyw: Oh, good catch. Not sure why we need the hard link, if the final resting place of the file, as returned from this function, includes the version? This mv and rmdir is just to clean up the tmpdir. See most recent commit for a fix for this issue, does that seem reasonable?

jsirois (approved):
Yeah, I think this probably works.

I'm never quite sure on the atomicity of `mv` on various OSs / distros. I'm pretty positive an `ln -f` followed by a remove of the original structure is always atomic though.

benjyw:
Thanks for the eagle-eyed reviews and great feedback! Phew...

PR #98, opened 2021-05-27
allow prepending args to Pants invocation

tdyas:
Allow prepending arguments to a Pants invocation which is helpful for running Pants under a profiler.

### DIFF
--- a/pants
+++ b/pants
@@ -279,5 +279,5 @@ if [[ -n "${PANTS_SHA:-}" ]]; then
 fi
 
 # shellcheck disable=SC2086
-exec "${pants_python}" "${pants_binary}" ${pants_extra_args} \
+exec ${PANTS_PREPEND_ARGS:-} "${pants_python}" "${pants_binary}" ${pants_extra_args} \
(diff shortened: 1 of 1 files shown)


jsirois (approved):
I guess. But really just using this seems easy enough for anyone that has the wherewithal to be profiling pants:
```
foo ~/.cache/pants/setup/bootstrap-Linux-x86_64/2.5.1rc0_py38/bin/python ~/.cache/pants/setup/bootstrap-Linux-x86_64/2.5.1rc0_py38/bin/pants ...
```

You're seriously in the weeds at that point anyway.

tdyas:
It was easier to have this for me, then having to figure out what the invocation was especially with the venv. This is based on a version that I did for `pants` itself. This would mostly be good if we want users to run it under profile, then they wouldn't have to figure out the invocation.

Eric-Arellano (commented):
Yeah, +1 to John. I won't block, but agree with keeping this script simpler for the average user.

tdyas:
okay will close then

PR #99, opened 2021-06-22
Isolate virtualenv with Pex.

jsirois:
Use Pex to download and run virtualenv to work around isolation
issues with the virtualenv zipapp as outlined in:
+ https://github.com/pypa/virtualenv/issues/1873
+ https://github.com/pypa/virtualenv/issues/2133

### DIFF
--- a/pants
+++ b/pants
@@ -263,7 +263,7 @@ function bootstrap_virtualenv {
       staging_dir=$(tempdir "${PANTS_BOOTSTRAP}")
       cd "${staging_dir}"
       echo "${VIRTUALENV_REQUIREMENTS}" > requirements.txt
-      "${python}" "${pex_path}" -r requirements.txt -c virtualenv -o virtualenv.pex --seed --venv
+      "${python}" "${pex_path}" -r requirements.txt -c virtualenv -o virtualenv.pex
       mkdir -p "$(dirname "${bootstrapped}")"
       mv -f "${staging_dir}/virtualenv.pex" "${bootstrapped}"
       rm -rf "${staging_dir}"
--- a/pants
+++ b/pants
@@ -34,10 +34,24 @@ fi
 
 PANTS_BOOTSTRAP="${PANTS_SETUP_CACHE}/bootstrap-$(uname -s)-$(uname -m)"
 
-VIRTUALENV_ZIPAPP=virtualenv.pyz
+PEX_VERSION=2.1.42
+PEX_URL="https://github.com/pantsbuild/pex/releases/download/v${PEX_VERSION}/pex"
+PEX_EXPECTED_SHA256="69d6b1b1009b00dd14a3a9f19b72cff818a713ca44b3186c9b12074b2a31e51f"
+
 VIRTUALENV_VERSION=20.4.7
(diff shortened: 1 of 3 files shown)


tests/test_smoke_tests.py
```diff
@@ -50,7 +50,7 @@ def _maybe_run_pyenv_local(self, python_version: Optional[str]) -> Iterator[None
             return
 
         def is_compatible(pyenv_version: str) -> bool:
-            major, minor, _ = pyenv_version.split(".")
+            major, minor = pyenv_version.split(".")[:2]
```
jsirois, line 53: N.B.: The old logic blew up if you had a diverse pyenv like so:
```
$ pyenv versions --bare
3.5.10
3.5.9
3.6.13
3.6.8
3.7.10
3.8.1
3.8.9
pypy2.7-7.3.1
pypy3.6-7.3.3
pypy3.7-7.3.3
```

tox.ini
```diff
@@ -52,6 +52,8 @@ commands =
 [testenv:typecheck]
 deps =
     mypy
```
jsirois, line 54: Of course mypy should be pinned but I've got enough yaks atm and the error output told me what extra deps to add.

jsirois:
I've filed https://github.com/pypa/virtualenv/issues/2133 and pointed back here. Maybe virtualenv will move from their bespoke zipapp packaging to PEX. Until then though.

pants
```diff
@@ -34,10 +34,24 @@ fi
...
+PEX_VERSION=2.1.42
+PEX_URL="https://github.com/pantsbuild/pex/releases/download/v${PEX_VERSION}/pex"
+PEX_EXPECTED_SHA256="69d6b1b1009b00dd14a3a9f19b72cff818a713ca44b3186c9b12074b2a31e51f"
+
 VIRTUALENV_VERSION=20.4.7
-VIRTUALENV_ZIPAPP_URL="https://raw.githubusercontent.com/pypa/get-virtualenv/${VIRTUALENV_VERSION}/public/${VIRTUALENV_ZIPAPP}"
-VIRTUALENV_ZIPAPP_EXPECTED_SHA256="9d5c8b09cdeaf44414f07902d2ae45c4bb8a2085717d66f82ae389a9c6f94425"
+VIRTUALENV_REQUIREMENTS=$(
+cat << EOF
+virtualenv==${VIRTUALENV_VERSION} --hash sha256:2b0126166ea7c9c3661f5b8e06773d28f83322de7a3ff7d06f0aed18c9de6a76
```
jsirois, line 44: Alternatively, we could just build and host out own virtualenv.pex to retain single file download. As this PR stands, a shop with a private PyPI might have a lot more work to do to get this script working with their setup.

benjyw: Not a bad idea, although then that private PyPI shop would have to host that virtualenv.pex?

benjyw (approved):

pants
benjyw: Although I guess now they have to host pex.pex which has the same implications.

jsirois: Yeah, this is an existing problem with the virtualenv.pyz, the problem just flips to the Pex PEX. I was pointing out the problem expands a bit though, since PyPI (or some private index) gets involved in building the virtualenv.pex whereas it did not get involved before.

PR #102, opened 2021-08-18
Pants detects `python` and `python3` binaries.

### LINKED ISSUE #100: determine_default_python_exe should check the internal interpreter version instead of its executable name
I'm running python on macOS and have installed python3.9 as well as python3.8 using homebrew. The app I'm currently using pants for uses 3.8 whereas the default is now 3.9 (and many apps require that). Since I don't want to unlink and re-link 3.8 every time I do a brew upgrade, I keep 3.9 as the default and tried to setup my environment for said app to use a virtualenv with 3.8 instead.

For most tools & IDEs this was pretty easy, but the pants wrapper script doesn't seem to respect that version. Although `python` and `python3` link to the correct python3.8 binary (`/usr/local/Cellar/python@3.8/3.8.11/bin/python3`), the wrapper is hard coded to iterate through executables that are named `pythonX.X`. And since `virtualenv` doesn't create a link or shim in that format, I don't have a `python3.8` executable. I _do_ however have a `python3.9` executable (linked by homebrew), which then (wrongfully) becomes pants next best guess (despite having set `interpreter_constraints = ["==3.8.*"]`)*.

In my opinion, if `interpreter_search_paths = ["<PATH>"]` is set, pants should also consider `python` / `python3`, check their versions and see if they satisfy the configured constraints.  Moreover, it should rather check the _actual_ interpreter version instead of its bare name.

*I believe this might be a bug that is already resolved by now. The wrapper we're using was created sometime around pants 2.2 or 2.3

jsirois:
Previously the pants script only worked with `python<X>.<Y>` binaries.

Fixes #100

### DIFF
--- a/tests/test_smoke_tests.py
+++ b/tests/test_smoke_tests.py
@@ -2,15 +2,17 @@
 # Licensed under the Apache License, Version 2.0 (see LICENSE).
 
 """Test that `./pants` works correctly with some basic sanity checks."""
-
+import dataclasses
 import os
+import shlex
 import shutil
 import subprocess
+import tempfile
 import textwrap
 from contextlib import contextmanager
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Iterator, List, Optional
+from typing import Any, Iterable, Iterator, List, Mapping, Optional
 
 import pytest
 from helpers import create_pants_config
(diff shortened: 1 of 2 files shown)


tests/test_smoke_tests.py
```diff
@@ -66,16 +92,110 @@ def is_compatible(pyenv_version: str) -> bool:
...
+                        "/usr/bin/env",
+                        "bash",
+                        "-c",
+                        f"command -v {' '.join(shlex.quote(exe_name) for exe_name in exe_names)}",
+                    ],
+                    check=not fallible,
+                    stdout=subprocess.PIPE,
+                )
+                .stdout.decode()
+                .splitlines(),
```
Eric-Arellano, line 120: Would `shutil.which()` work? https://docs.python.org/3/library/shutil.html#shutil.which

Eric-Arellano (approved):
Thanks for improving this

tests/test_smoke_tests.py
jsirois: Calling it 20+ times would, but not once like this to get everything.

PR #104, opened 2021-09-09
update default branch name in pants script

asherf:
the default branch on the pants repo is `main` not `master` 
Also I am assuming this is the script hat gets published to https://static.pantsbuild.org/setup/pants
if this is the case, this update is worth doing and re-publishing (I assume there is a manual step since the script in https://static.pantsbuild.org/setup/pants is not the same one in this repo... just very similar)

### DIFF
--- a/pants
+++ b/pants
@@ -12,8 +12,8 @@
 
 set -eou pipefail
 
-# NOTE: To use an unreleased version of Pants from the pantsbuild/pants master branch,
-#  locate the master branch SHA, set PANTS_SHA=<SHA> in the environment, and run this script as usual.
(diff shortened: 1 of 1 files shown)


Eric-Arellano (approved):

benjyw (approved):

benjyw:
https://static.pantsbuild.org/setup/pants just redirects to https://pantsbuild.github.io/setup/pants so there shouldn't be a manual step needed. 

```
$ curl -i https://static.pantsbuild.org/setup/pants
HTTP/2 302 
content-length: 0
location: https://pantsbuild.github.io/setup/pants
```

PR #109, opened 2021-09-22
Switch from Travis-CI to GH Actions.

jsirois:


stuhood (approved):
Pre-approving. Thanks a lot for doing this!

benjyw (approved):
Nice! Thanks for doing this.

jsirois:
@Eric-Arellano I wanted your take on ditching pyenv. It's not needed for CI anymore and it made both CI setup and the test code bigger / more complicated. You can still use pyenv as a dev (I do), you just need to make sure you've arranged that the needed Python versions are on your $PATH.

jsirois:
I'm going to proceed with this @Eric-Arellano but I'll follow up if you point out something I've missed with ditching pyenv.

Eric-Arellano (commented):
Sounds good on dropping Pyenv! I had thought we had a test for our Pyenv shim handling, but looks like not. That was the only thing I would have been worried about.

PR #110, opened 2022-01-05
silence asdf warnings from ./pants for python 3.9

ryanking:
If you are using [asdf](http://asdf-vm.com/) to manage python versions
and don't have 3.9 installed+configured, each run of pants will show an
error like:

```
No preset version installed for command python3.9
Please install a version by running one of the following:

asdf install python 3.6.15
asdf install python 3.8.12

or add one of the following versions in your config file at /Users/ryan.king/code/color/.tool-versions
python 3.9.1
python 3.9.7
```

The script will continue on, but I think this error would be confusing
for many people and this change should be safe.

### DIFF
--- a/pants
+++ b/pants
@@ -191,12 +191,8 @@ function determine_default_python_exe {
     if [[ -z "${interpreter_path}" ]]; then
       continue
     fi
-    # Check if the Python version is installed via Pyenv but not activated.
-    if [[ "$("${interpreter_path}" --version 2>&1 > /dev/null)" == "pyenv: python${version}"* ]]; then
-      continue
-    fi
-    # Check if a version is shimmed by asdf but not configured.
-    if ! ${interpreter_path} --version >/dev/null 2>&1; then
+    # Check if a version is shimmed by pyenv or asdf but not configured.
+    if ! "${interpreter_path}" --version >/dev/null 2>&1; then
       continue
     fi
     if [[ -n "$(check_python_exe_compatible_version "${interpreter_path}")" ]]; then
--- a/pants
+++ b/pants
@@ -195,6 +195,10 @@ function determine_default_python_exe {
     if [[ "$("${interpreter_path}" --version 2>&1 > /dev/null)" == "pyenv: python${version}"* ]]; then
       continue
     fi
+    # Check if a version is shimmed by asdf but not configured.
+    if ! ${interpreter_path} --version >/dev/null 2>&1; then
+      continue
+    fi
     if [[ -n "$(check_python_exe_compatible_version "${interpreter_path}")" ]]; then
       echo "${interpreter_path}" && return 0
     fi


benjyw:
Thanks for the contribution! I'm of two minds about this, overall I think this makes sense. But I do worry about it making it harder to debug interpreter issues (by squelching output that doesn't truly indicate an error the user needs to care about we're also squelching output that does). 

What is the underlying cause of these warnings? Does `asdf` set up shims for python versions that aren't installed?

Eric-Arellano:
This is how we approach it for Pyenv:

https://github.com/pantsbuild/setup/blob/3ef18abce129b5c07556973b8e19aba4d13336c2/pants#L187-L202

I wonder if we could do something similar here?

ryanking:
> What is the underlying cause of these warnings? Does `asdf` set up shims for python versions that aren't installed?

Yes. Maybe I can find a way to un-shim python3.9.

ryanking:
@Eric-Arellano I believe that asdf uses pyenv under the hood, so something similar might work

ryanking:
@Eric-Arellano @benjyw does this make more sense? An interpretter shimmed by asdf but not configured will exit `126` when invoked. This might make sense to use with pyenv too?

benjyw:
> @Eric-Arellano @benjyw does this make more sense? An interpretter shimmed by asdf but not configured will exit `126` when invoked. This might make sense to use with pyenv too?

Looks like an unpointed pyenv shim exits `127`, so yeah, this could work as a general solution. Nice!

pants
```diff
@@ -195,6 +195,10 @@ function determine_default_python_exe {
     if [[ "$("${interpreter_path}" --version 2>&1 > /dev/null)" == "pyenv: python${version}"* ]]; then
```
benjyw: Looks like we can remove this check in favor of the new one!

pants
```diff
@@ -195,6 +195,10 @@ function determine_default_python_exe {
...
       continue
     fi
+    # Check if a version is shimmed by asdf but not configured.
+    if ! ${interpreter_path} --version >/dev/null 2>&1; then
```
benjyw: My shell-fu is not strong, but I think this needs quoting, in case the path has spaces in it?

ryanking:
Updated to resolve feedback.

Also note that is appears shellfmt is not run on this script, which makes it a bit annoying to work on with en editor configured to use shellfmt. :)

benjyw:
Good point. @Eric-Arellano could the `./pants` script itself be a source for a `shell_sources` target? :)

Eric-Arellano:
> Good point. @Eric-Arellano could the ./pants script itself be a source for a shell_sources target? :)

Yeah. We switched from using Pants for CI to Tox in https://github.com/pantsbuild/setup/pull/64. But we could/probably should go back to using Pants. 

A major motivation in that PR was getting rid of those build-support scripts -- I don't think we'd need those anymore because Pants has gotten much better. For example, we now have Shellcheck and Shfmt.

Eric-Arellano:
> An interpretter shimmed by asdf but not configured will exit 126 when invoked. This might make sense to use with pyenv too?

Thanks for continuing to iterate on this! I might be missing something, but I don't see anything about exit codes 126 and 127 in the PR. It looks like this will squash all errors?

benjyw (approved):
Thanks! Having less code but more functionality seems like a win to me!

Will let @jsirois weigh in just in case.

benjyw:
Squashing all errors is probably the right thing to do? If an interpreter is unusable for our purposes we should just ignore it no matter what? The purpose of this script is to run Pants, not perform interpreter diagnostics...

Eric-Arellano:
> Squashing all errors is probably the right thing to do?

Yeah, that's a good point! And it's a bash script users can edit if they want to debug.

jsirois (approved):

benjyw:
Thanks @ryanking !! This is great.

Issue #116, opened 2022-01-31
Unexpected use of HTTP server in Pants script download flow


christophermaier:
I was trying to automate some Pants updates and was looking to use this invocation to update the pants script itself:
```
curl --proto "=https" \
    --tlsv1.3 \
    --location \
    --verbose \
    --output ./pants \
    https://pantsbuild.github.io/setup/pants
```
However, when I run that, it appears that at some point in the flow, I get redirected to an http site:
```
HTTP/2 301 
< server: GitHub.com
< content-type: text/html
< permissions-policy: interest-cohort=()
< location: http://v1.pantsbuild.org/setup/pants
< x-github-request-id: 4DDA:0AD7:24AE8:607545:61F8148F
< accept-ranges: bytes
< date: Mon, 31 Jan 2022 16:57:01 GMT
< via: 1.1 varnish
< age: 78
< x-served-by: cache-fty21360-FTY
< x-cache: HIT
< x-cache-hits: 1
< x-timer: S1643648221.348955,VS0,VE1
< vary: Accept-Encoding
< x-fastly-request-id: de662d2ef47f018f97d6677537bf7d7ea081cf13
< content-length: 162
< 
* Ignoring the response-body
{ [162 bytes data]
100   162  100   162    0     0   2793      0 --:--:-- --:--:-- --:--:--  2842
* Connection #0 to host pantsbuild.github.io left intact
* Issue another request to this URL: 'http://v1.pantsbuild.org/setup/pants'
* Protocol "http" not supported or disabled in libcurl
* Closing connection -1
```

benjyw:
Not sure if this is related/will fix, but the officially supported URL to download is `https://static.pantsbuild.org/setup/pants`

christophermaier:
I tried with the new URL and it behaved in the same manner, unfortunately.

benjyw:
This is related to configuration of custom GH Pages domains. 

We have a slightly complicated situation: 

The Pants v1 docs are statically served on GH Pages at https://pantsbuild.github.io/ from a dedicated repo (https://github.com/pantsbuild/pantsbuild.github.io/). But we also have second repo (https://github.com/pantsbuild/setup/) that serves GH Pages at a subpath https://pantsbuild.github.io/setup, just for the setup script. This is because GH Pages have to be served under `<ORG>.github.io`, you only get one subdomain per org.

[...]

PR #117, opened 2022-02-01
Enforce use of HTTPS with >= TLS 1.2 for curl calls

christophermaier:
This is a follow-on to https://github.com/pantsbuild/setup/issues/116,
tightening up the security of the `curl` calls made in the `pants`
script.

I also took the liberty of replacing single-character options with
their long option equivalents for readability.

Signed-off-by: Christopher Maier <chris@graplsecurity.com>

### DIFF
--- a/pants
+++ b/pants
@@ -254,7 +254,11 @@ function bootstrap_pex {
       local staging_dir
       staging_dir=$(tempdir "${PANTS_BOOTSTRAP}")
       cd "${staging_dir}"
-      curl -LO "${_PEX_URL}"
+      curl --proto "=https" \
+           --tlsv1.3 \
+           --location \
+           --remote-name \
+           "${_PEX_URL}"
       fingerprint="$(compute_sha256 "${python}" "pex")"
       if [[ "${_PEX_EXPECTED_SHA256}" != "${fingerprint}" ]]; then
         die "SHA256 of ${_PEX_URL} is not as expected. Aborting."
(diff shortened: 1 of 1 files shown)


benjyw:
Thanks for this! 

My one concern is the `--tlsv1.3` flag. On my macbook it makes curl fails with this error: `curl: (4) LibreSSL was built without TLS 1.3 support`

So I fear widespread breakage on systems where curl does not support TLS 1.3 yet.

benjyw:
Maybe my system is unusual in some way? I'd be interested in seeing if this is a common error on macbooks.

christophermaier:
@benjyw No, your system isn't unusual... I think mine is 😅 I just ran into a TLS failure along these same lines when a similar change I was making on a work project failed in our CI, which apparently doesn't have TLS 1.3 either. I'm doing some further testing on that codebase to land on the right combination of options to make this more broadly accessible; I'll update this PR once I get something working.

Sorry for the noise! 🙇‍♂️

christophermaier:
I've relaxed things to TLS 1.2. I'm going to do a bit more research to see if there's a sane way to support folks that may not even have TLS 1.2 support in their `curl` binaries.

(I left off TLS 1.0 and 1.1 since they're deprecated, but depending on how you want to go, I could relax things even more.)

benjyw:
Thanks! I think TLS v1.2 is fine to assume - that flag was added to curl 9 in 2013, and TLS 1.2 itself is from 2008.

christophermaier:
@benjyw Sounds good to me. If 1.2 is fine with you, maybe we can just leave it there. In that case, I think I'm done with changes to this PR.

Thanks!

benjyw:
Yeah, I think it's fine to require TLS 1.2. If someone really doesn't have support for that, they can always edit the file, this is just the default. 

Thanks for the contribution!

benjyw (approved):

Eric-Arellano:
> I think it's fine to require TLS 1.2

Not only fine, but desirable imo. I'm glad for Pants to play a (small) part in making computing more secure. Thank you for your contribution!

If this is a real issue for anyone, they can change the bash script to 1.1.

Eric-Arellano (approved):

christophermaier:
(Updated the PR title for the benefit of your changelog!)

PR #119, opened 2022-02-23
Make pants bootstrap process produce less output and be less verbose

asherf:
I feel that the current output generated by the script is not very useful to most users (who just run pants in their repo, and not managing it)
so I think pants should have a cleaner and simpler UI and a clean bootstrap process is part of it.

@jsirois @stuhood @benjyw @Eric-Arellano

### DIFF
--- a/pants
+++ b/pants
@@ -256,6 +256,7 @@ function bootstrap_pex {
       cd "${staging_dir}"
       curl --proto "=https" \
            --tlsv1.2 \
+           --silent \
            --location \
            --remote-name \
            "${_PEX_URL}"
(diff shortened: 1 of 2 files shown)


jsirois:
I'll let others review this. I'm against, but I suspect I have the wrong sensibilities here.

pants
```diff
@@ -356,11 +357,11 @@ function bootstrap_pants {
       (
         scrub_PEX_env_vars
         # shellcheck disable=SC2086
-        "${python}" "${virtualenv_path}" --no-download "${staging_dir}/install" && \
+        "${python}" "${virtualenv_path}" --quiet --no-download "${staging_dir}/install" && \
```
stuhood, line 360: Does `--quiet` still render all of the output on failure? If not, it might be better to pipe it to a temporary log, and cat the log on failure of this block.

stuhood (commented):
When we finally get around to https://github.com/pantsbuild/pants/issues/12397, it will be relevant here too, since it will eliminate the PIP steps.

pants
asherf: errors still show up.
Screen Shot 2022-02-23 at 7 39 08 PM

Eric-Arellano (approved):
+1 for me. I agree with Asher that most people don't care about this info.

I'll wait to merge till there's more consensus though.

benjyw (approved):
Great stuff!

PR #121, opened 2022-04-04
Scrub PYTHONPATH in addition to PEX env vars

### LINKED ISSUE #120: Bootstrap script won't install dependencies pip finds in current env
For complex reasons, `./pants` is being invoked as a subprocess of some other Python process. If I wipe the bootstrap dir (and get rid of `--quiet` from the relevant `pip install` in the bootstrap script) I see:

```
Requirement already satisfied: typing-extensions==4.0.1 in /path/to/typing_extensions (from pantsbuild.pants==2.11.0rc1) (4.0.1)
...
```

for > 1 requirements. Then if I invoke Pants **outside** the subprocess I'll get an import error for those packages.

Should pants use the following pip option?

```
 -I, --ignore-installed      Ignore the installed packages, overwriting them. This can break your system if the existing package is of a different version or was
                              installed with a different package manager!
```

(which works in my use case)

thejcannon:
Fixes https://github.com/pantsbuild/setup/issues/120

### DIFF
--- a/pants
+++ b/pants
@@ -282,7 +282,7 @@ function scrub_env_vars {
   fi
   # Also ensure pip doesn't think packages on PYTHONPATH
   # are already installed.
-  if [[ -n "$PYTHONPATH" ]]; then
+  if [[ -n "${PYTHONPATH}" ]]; then
     warn "Scrubbing PYTHONPATH"
     unset "PYTHONPATH"
   fi
--- a/pants
+++ b/pants
@@ -282,7 +282,7 @@ function scrub_env_vars {
   fi
   # Also ensure pip doesn't think packages on PYTHONPATH
   # are already installed.
-  if [ -n '$PYTHONPATH' ]; then
+  if [ -n "$PYTHONPATH" ]; then
     warn "Scrubbing PYTHONPATH"
     unset PYTHONPATH
   fi
--- a/pants
+++ b/pants
@@ -282,7 +282,7 @@ function scrub_env_vars {
   fi
   # Also ensure pip doesn't think packages on PYTHONPATH
   # are already installed.
-  if [ -n $PYTHONPATH ]; then
+  if [ -n '$PYTHONPATH' ]; then
     warn "Scrubbing PYTHONPATH"
     unset PYTHONPATH
   fi
--- a/pants
+++ b/pants
@@ -273,12 +273,18 @@ function bootstrap_pex {
   echo "${bootstrapped}"
 }
 
-function scrub_PEX_env_vars {
+function scrub_env_vars {
   # Ensure the virtualenv PEX runs as shrink-wrapped.
   # See: https://github.com/pantsbuild/setup/issues/105
   if [[ -n "${!PEX_@}" ]]; then
     warn "Scrubbing ${!PEX_@}"
     unset "${!PEX_@}"
+  fi
+  # Also ensure pip doesn't think packages on PYTHONPATH
+  # are already installed.
+  if [[ -n "$PYTHONPATH" ]]; then
+    warn "Scrubbing PYTHONPATH"
+    unset PYTHONPATH
   fi
 }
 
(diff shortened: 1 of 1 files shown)


asherf (approved):

asherf:
@jsirois @Eric-Arellano should also review.

pants
```diff
@@ -361,7 +361,7 @@ function bootstrap_pants {
         # Grab the latest pip, but don't advance setuptools past 58 which drops support for the
         # `setup` kwarg `use_2to3` which Pants 1.x sdist dependencies (pystache) use.
         "${staging_dir}/install/bin/pip" install --quiet -U pip "setuptools<58" && \
-        "${staging_dir}/install/bin/pip" install ${maybe_find_links} --quiet --progress-bar off "${pants_requirement}"
+        "${staging_dir}/install/bin/pip" install ${maybe_find_links} --ignore-installed --quiet --progress-bar off "${pants_requirement}"
```
jsirois: This doesn't make sense even if it works. The venv this is running in was created fresh 2 non-comment lines above and we are not using `--system-site-packages` when creating that fresh venv. As such, there should be nothing installed - except `pip`, `setuptools` and `wheel`.

That gives me pause. Do you understand how this works - aka why pip is seeing anything as installed in this fresh venv at all? Do you have things like PYTHONPATH set in your execution environment that are modifying what might be seen to include things outside this fresh venv perhaps?

At the very least a comment explaining what's going on here if you understand it (I don't). Perhaps also the less dangerous `--upgrade --upgrade-strategy eager` if that works.

thejcannon: > Do you have things like PYTHONPATH set in your execution environment 

Yup!

> Perhaps also the less dangerous --upgrade --upgrade-strategy eager if that works.

It has the same issue.

thejcannon: There's `--isolated` but that ignores user config which I don't think is strictly-speaking what we want.

thejcannon: This is easily reproducible, will add to issue.

thejcannon: https://github.com/pantsbuild/setup/issues/120#issuecomment-1087995877

jsirois: OK, thanks for checking. How about amending scrub_PEX_env_vars to also scrub PYTHONPATH?

jsirois: OR, if you stick with the current, please just add a comment saying how the heck there can be anything installed to ignore - because its not installed, but it is on some ambient PYTHONPATH.

thejcannon: Scrubbing the env is likely the right approach.
I'll try it out tomorrow.

jsirois (approved):

thejcannon:
Sorry for spam, I've been editing this in browser and am no good at Bash.

kaos (approved):

pants
```diff
@@ -282,7 +282,7 @@ function scrub_env_vars {
   fi
   # Also ensure pip doesn't think packages on PYTHONPATH
   # are already installed.
-  if [ -n $PYTHONPATH ]; then
+  if [ -n '$PYTHONPATH' ]; then
```
kaos: Good idea to quote, but you want to use double quotes..

```bash
$ echo "$USER"
myname
$ echo '$USER'
$USER
```

thejcannon: OK then I have no idea how to not make this line complain about unbound local. Any Bash folks want to help? I swear this should be easy.

kaos: try "${PYTHONPATH:-}" then :)

jsirois: Use: `"${PYTHONPATH:-}"`, That says, PYTHONPATH or default to the thing after `-`, which is the empty string.

jsirois: Handy: https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

thejcannon: Thanks y'all.

Eric-Arellano: This reminds me that we should really be using shellcheck on the script. It would require dogfooding pants which we used to do, but I removed it two years ago which I now regret

Eric-Arellano (approved):

PR #123, opened 2022-05-20
Source PANTS_BOOTSTRAP script, defaults to .pants.bootstrap, if it exists.

kaos:
This is a change we (at $-work) apply to all our `pants` scripts to setup defaults for running locally, values that are otherwise populated when running in CI/CD, as well as setting up some custom cert bundles etc.

It helps ensuring a proper environment to run pants in, without having to do any pre-steps to invoking `./pants`.

### DIFF
--- a/pants
+++ b/pants
@@ -12,6 +12,12 @@
 
 set -eou pipefail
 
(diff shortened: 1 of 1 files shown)


benjyw (approved):
Nice!

PR #124, opened 2022-06-01
Adds one-step setup script for Pants setup.

chrisjrn:
Closes [pantsbuild/#15526](https://github.com/pantsbuild/pants/issues/15526)

### DIFF
--- a/one_step_setup
+++ b/one_step_setup
@@ -1,0 +1,61 @@
+#!/usr/bin/env bash
+
+set -e
+
+function fail {
+    # Print a newline to stderr and exit.
+    echo > /dev/stderr
+    exit 1
+}
+
+
+# `$PANTS_VERSION` must be set
+if [ -z "$PANTS_VERSION" ]
+then
+    echo \`\$PANTS_VERSION\` must be set. Finish installing Pants by re-running this script \
+        with \`PANTS_VERSION\` set. > /dev/stderr
+    fail
+fi
+
+# Make sure we don't override an existing Pants installation.
+if [ -f "pants.toml" ]
+then
+    echo A \`pants.toml\` file already exists in this directory. This installation script \
+        needs to create a new \`pants.toml\` file. > /dev/stderr
+    echo If you are trying to upgrade to a newer version of Pants, set the \`pants_version\` \
+        setting in your existing \`pants.toml\` file. > /dev/stderr
+    echo If you are trying to run the install process from scratch, delete the \`pants.toml\` \
+        file and start again. > /dev/stderr
+    fail 
+fi
+
+if [ -f "pants" ]
+then
+    echo A file called \`pants\` already exists in this directory. This installation script \
+        will overwrite this file. To proceed, either rename this file, or delete the \
+        \`pants\` file. > /dev/stderr
+    fail
+fi
+
+# Create enough of a pants.toml file that our bootstrap process can run
+printf '[GLOBAL]\npants_version = "'$PANTS_VERSION'"\n' > pants.toml
+
+# Fetch the Pants bootstrap script and run it to verify that we fetched
+# the correct version
+curl -L -O https://static.pantsbuild.org/setup/pants 
+chmod +x ./pants
+PANTS_EXEC_VERSION=`./pants --version`
+
+# Verify that the correct version of Pants was installed.
+if [ $PANTS_VERSION != $PANTS_EXEC_VERSION ]
+then
+    echo Pants was installed, but the version was $PANTS_EXEC_VERSION, not \
+        $PANTS_VERSION. > /dev/stderr
+    echo > /dev/stderr
+    fail
+fi
+
+# Let the user know that everything worked
+echo > /dev/stderr
+echo Pants was installed successfully! > /dev/stderr
+echo > /dev/stderr
--- a/one_step_setup.sh
+++ b/one_step_setup.sh
@@ -47,7 +47,7 @@ if [ $PANTS_VERSION != $PANTS_EXEC_VERSION ]
 then
     echo Pants was installed, but the version was $PANTS_EXEC_VERSION, not \
         $PANTS_VERSION. > /dev/stderr
-    echo > /dev/stderr
+    echo Consider filing an issue at https://github.com/pantsbuild/setup/issues/new > /dev/stderr
     fail
 fi
 
--- a/one_step_setup.sh
+++ b/one_step_setup.sh
@@ -28,10 +28,13 @@ then
     fail
 fi
 
-# Find the latest stable version from PyPI
-PANTS_VERSION=`curl https://pypi.org/pypi/pantsbuild.pants/json | \
-    grep -o '"version":"[^"]*"' | \
-    grep -o "[0-9]*\\.[0-9]*\\.[0-9]*"`
+# Find the latest stable version from PyPI if not set at the command line.
+if [ -z "$PANTS_VERSION" ]
+then
+    PANTS_VERSION=`curl https://pypi.org/pypi/pantsbuild.pants/json | \
+        grep -o '"version":"[^"]*"' | \
+        grep -o "[0-9]*\\.[0-9]*\\.[0-9]*"`
+fi
 
 # Create enough of a pants.toml file that our bootstrap process can run
 printf '[GLOBAL]\npants_version = "'$PANTS_VERSION'"\n' > pants.toml
--- a/one_step_setup.sh
+++ b/one_step_setup.sh
@@ -8,15 +8,6 @@ function fail {
     exit 1
 }
 
-
-# `$PANTS_VERSION` must be set
-if [ -z "$PANTS_VERSION" ]
-then
-    echo \`\$PANTS_VERSION\` must be set. Finish installing Pants by re-running this script \
-        with \`PANTS_VERSION\` set. > /dev/stderr
-    fail
-fi
-
 # Make sure we don't override an existing Pants installation.
 if [ -f "pants.toml" ]
 then
@@ -36,6 +27,11 @@ then
         \`pants\` file. > /dev/stderr
     fail
 fi
+
+# Find the latest stable version from PyPI
+PANTS_VERSION=`curl https://pypi.org/pypi/pantsbuild.pants/json | \
+    grep -o '"version":"[^"]*"' | \
+    grep -o "[0-9]*\\.[0-9]*\\.[0-9]*"`
 
 # Create enough of a pants.toml file that our bootstrap process can run
 printf '[GLOBAL]\npants_version = "'$PANTS_VERSION'"\n' > pants.toml
@@ -57,5 +53,5 @@ fi
 
 # Let the user know that everything worked
 echo > /dev/stderr
-echo Pants was installed successfully! > /dev/stderr
+echo Pants was installed successfully\! > /dev/stderr
 echo > /dev/stderr
--- a/tests/test_one_step_setup.py
+++ b/tests/test_one_step_setup.py
@@ -1,0 +1,50 @@
+# Copyright 2019 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+"""Test that the first time installation flow described by
+https://www.pantsbuild.org/install.html#recommended-installation works as expected."""
+
+import os
+import re
+import subprocess
+from pathlib import Path
+
+
+def test_runs_on_clean_directory(tmp_path: Path) -> None:
+
+    cwd = os.getcwd()
+
+    proc = subprocess.run(
+        ["/bin/bash", os.path.join(cwd, "one_step_setup.sh")],
+        cwd=tmp_path,    
+    )
+
+    assert proc.returncode == 0
+
+
+def test_fails_if_pants_script_present(tmp_path: Path) -> None:
+
+    cwd = os.getcwd()
+    pants_script = tmp_path / "pants"
+    pants_script.touch()
+
+    proc = subprocess.run(
+        ["/bin/bash", os.path.join(cwd, "one_step_setup.sh")],
+        cwd=tmp_path,    
+    )
+
+    assert proc.returncode == 1
+
+
+def test_fails_if_pants_toml_present(tmp_path: Path) -> None:
+
+    cwd = os.getcwd()
+    pants_toml = tmp_path / "pants.toml"
+    pants_toml.touch()
+
+    proc = subprocess.run(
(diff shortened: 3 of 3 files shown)


one_step_setup
```diff
@@ -0,0 +1,61 @@
...
+# the correct version
+curl -L -O https://static.pantsbuild.org/setup/pants 
+chmod +x ./pants
+PANTS_EXEC_VERSION=`./pants --version`
+
+# Verify that the correct version of Pants was installed.
+if [ $PANTS_VERSION != $PANTS_EXEC_VERSION ]
+then
+    echo Pants was installed, but the version was $PANTS_EXEC_VERSION, not \
+        $PANTS_VERSION. > /dev/stderr
```
Eric-Arellano: Maybe mention filing a bug at this repo.

one_step_setup
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env bash
+
+set -e
+
+function fail {
+    # Print a newline to stderr and exit.
+    echo > /dev/stderr
```
Eric-Arellano: Why `/dev/stderr` rather than the number? 1 iirc? Not a shell fan.

one_step_setup
```diff
@@ -0,0 +1,61 @@
...
+then
+    echo Pants was installed, but the version was $PANTS_EXEC_VERSION, not \
+        $PANTS_VERSION. > /dev/stderr
+    echo > /dev/stderr
+    fail
+fi
+
+# Let the user know that everything worked
+echo > /dev/stderr
+echo Pants was installed successfully! > /dev/stderr
```
Eric-Arellano: Should this be quoted with the `!`?

one_step_setup
```diff
@@ -0,0 +1,61 @@
...
+if [ -f "pants" ]
+then
+    echo A file called \`pants\` already exists in this directory. This installation script \
+        will overwrite this file. To proceed, either rename this file, or delete the \
+        \`pants\` file. > /dev/stderr
+    fail
+fi
+
+# Create enough of a pants.toml file that our bootstrap process can run
+printf '[GLOBAL]\npants_version = "'$PANTS_VERSION'"\n' > pants.toml
```
Eric-Arellano: We probably want to add conditional logic to add certain defaults like `use_deprecated_cli_args = false`, based on the Pants version they're targeting. Fine in a followup, as long as we have a way to do it. `./pants` already had logic to parse version strings.

one_step_setup
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env bash
```
Eric-Arellano: Maybe add `.sh` to the file?

Eric-Arellano (commented):
I like it! What will the command be to run this?

--

Thoughts on how we would handle making this more interactive, like dynamically asking people which backends to activate?

--

Thoughts on writing this as a Python 3 script? I wonder if it would make this more maintainable when we want added functionality like which backends you want to activate.

We require Python 3 anyways to run Pants itself, so it's a reasonable expectation (for now at least).

--

It would be good to add an integration test that this works. And to run it with Shellcheck and Shfmt (possibly via Pants)

chrisjrn:
@Eric-Arellano Feedback on review comment:

> I like it! What will the command be to run this?

``` 
PANTS_VERSION=2.12.0 /bin/bash -c "$(curl -fsSL https://{PANTS_URL}/one_step_setup)"
```

which is still a bit awkward, but it's also copyable from docs, and will only take one step.

> Thoughts on how we would handle making this more interactive, like dynamically asking people which backends to activate?

I think we should add that within Pants, because it'll be easier to test and keep up to date (e.g. Pants itself will have a better understanding of what backends are available for configuration)

> Thoughts on writing this as a Python 3 script? I wonder if it would make this more maintainable when we want added functionality like which backends you want to activate.
> 
> We require Python 3 anyways to run Pants itself, so it's a reasonable expectation (for now at least).

At the moment, the Python interpreter checks are in the `./pants` bootstrap process, so we already have a robust check in place, which we'll eventually hit if we keep this as a very minimal shell script. If we make this a Py3 script, then we'll have to rely on the user reading docs to understand what needs to be in place.

> It would be good to add an integration test that this works. And to run it with Shellcheck and Shfmt (possibly via Pants)

I'll add another integration test in the style of the tests for `./pants` (they're Python tests run by tox). Do not particularly want to pantsify this repo just for the one step setup, unless we want to do the entire thing. Do we?

one_step_setup
chrisjrn: The fact that you're pointing out that you can't remember the alias for stderr is a good sign that it wouldn't be clear enough to the reader. The extra characters are worth the added clarity.

one_step_setup
chrisjrn: I'm ambivalent here.

one_step_setup
chrisjrn: I'd somewhat prefer for this to be implemented in Pants itself, if we can get away with it.

Eric-Arellano:
> I think we should add that within Pants, because it'll be easier to test and keep up to date (e.g. Pants itself will have a better understanding of what backends are available for configuration)

Okay, makes sense to me. So, is the thinking for now "this gets us good enough" and we aren't going to keep improving for the next N weeks? If so, totally reasonable. Hopefully other things will change like how we distribute Pants entirely.

> (they're Python tests run by tox)

Yes, sounds good

one_step_setup
Eric-Arellano: I don't see how, given the resistance a year ago to a `generate-pants-toml` goal.

one_step_setup
Eric-Arellano: I vaguely think `.sh` makes it feel less scary to `curl` and execute. Ack that there is _zero_ actual security soundness to what I just said. File extension doesn't matter. But, it at least gives an additional hint to the reader that this is a Bash script

one_step_setup
```diff
@@ -0,0 +1,61 @@
...
+    # Print a newline to stderr and exit.
+    echo > /dev/stderr
+    exit 1
+}
+
+
+# `$PANTS_VERSION` must be set
+if [ -z "$PANTS_VERSION" ]
+then
+    echo \`\$PANTS_VERSION\` must be set. Finish installing Pants by re-running this script \
```
benjyw: I think we should default to latest stable version. We can hard-code it, and add to the release instructions an item about updating it on stable releases. 

But we can also detect it with 

```
curl https://pypi.org/pypi/pantsbuild.pants/json | jq -r .info.version
```

benjyw (commented):
Cool stuff! 

So the idea is that you run this but don't check it in?

Also, once this is deployed remind me to set up Google Analytics for tracking it.

chrisjrn:
@benjyw The idea is that the script is ephemeral -- you either download it and run it, or you can use `bash -c "$(curl …)"` like Homebrew recommenDs.

one_step_setup
chrisjrn: Defaulting to the latest stable version makes sense. If we can figure out how to do this automatically without adding a dependency on `jq`, then I'm all for it.

The pypi JSON blob is annoyingly dense. But This approach seems like the beginning of something that could work:

```
% curl https://pypi.org/rss/project/pantsbuild-pants/releases.xml | grep title | grep -v "a[0-9]" | grep -v "dev[0-9]" | grep -v "rc[0-9]"
    <title>PyPI recent updates for pantsbuild.pants</title>
      <title>2.10.1</title>
      <title>2.9.2</title>
      <title>2.8.1</title>
      <title>2.11.0</title>
      <title>2.9.1</title>
      <title>2.10.0</title>
```

one_step_setup
chrisjrn: I think it might be a bit more welcome now. There's a good reason to do backend selection from within Pants (because it might be a bit more self-aware), for instance.

chrisjrn:
@Eric-Arellano 

> Okay, makes sense to me. So, is the thinking for now "this gets us good enough" and we aren't going to keep improving for the next N weeks? If so, totally reasonable. Hopefully other things will change like how we distribute Pants entirely.

I think so. This should be enough to put in the docs as-is.

one_step_setup
benjyw: This gets you close:

```
curl https://pypi.org/rss/project/pantsbuild-pants/releases.xml | grep -o '"version":"[^"]*"'
```

Or writing this in Python...

chrisjrn: @benjyw Nice. I can work with that.

chrisjrn: @benjyw This seems to work just fine:

```
PANTS_VERSION=`curl https://pypi.org/pypi/pantsbuild.pants/json  | grep -o '"version":"[^"]*"' | grep -o "[0-9]*\\.[0-9]*\\.[0-9]*"`
```

one_step_setup
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env bash
+
+set -e
+
```
stuhood: Thoughts on embedding all of this logic in the `pants` script instead?

How I could imagine that working would be that if no `pants.toml` existed, it would create one for you.

Not sure if that would run afoul of users using non-default config file names, but you could imagine disabling the behavior by looking at env vars.

chrisjrn: I think keeping the script simple is good for inspectability -- the intention is for people to execute directly from `curl`, so having the script be simpler is good.

one_step_setup
Eric-Arellano: As the person who took away this affordance in https://github.com/pantsbuild/setup/pull/86...I am okay with it because we are going to be pinning the version in `pants.toml`.

We probably want on dev versions to have people set the `PANTS_VERSION` explicitly though? So our docs still need to mention this env var?

one_step_setup
Eric-Arellano: +1 to having the dedicated script so that people can inspect what they're about to run.

tests/test_one_step_setup.py
```diff
@@ -0,0 +1,49 @@
...
+def test_runs_on_clean_directory(tmp_path: Path) -> None:
+
+    cwd = os.getcwd()
+
+    proc = subprocess.run(
+        ["/bin/bash", os.path.join(cwd, "one_step_setup.sh")],
+        cwd=tmp_path,
+    )
+
+    assert proc.returncode == 0
```
Eric-Arellano, line 21: maybe worth also asserting that stdout has something we want? deleting all the contents of this script would still cause this test to pause.

one_step_setup
chrisjrn: That seems like a good idea. I'll add support for setting the variable still.

Eric-Arellano (approved):
Looks great to me! 

(Given this is a new workflow, please wait for sign off from 1+ others)

benjyw (approved):
FWIW I prefer this as a separate script for the reasons stated, and also because we don't want to update the script ever user uses on every run just to modify functionality that is used exactly once by one user. 

Also, tracking the download analytics on this script will be more informative about numbers of repos evaluating or adopting pants than downloads of the `pants` script will be.

tests/test_one_step_setup.py
chrisjrn: Good idea. I've added that to the tests.

one_step_setup
chrisjrn: @Eric-Arellano @benjyw I've added this feature.

one_step_setup.sh
```diff
@@ -0,0 +1,60 @@
...
+        grep -o '"version":"[^"]*"' | \
+        grep -o "[0-9]*\\.[0-9]*\\.[0-9]*"`
+fi
+
+# Create enough of a pants.toml file that our bootstrap process can run
+printf '[GLOBAL]\npants_version = "'$PANTS_VERSION'"\n' > pants.toml
+
+# Fetch the Pants bootstrap script and run it to verify that we fetched
+# the correct version
+curl -L -O https://static.pantsbuild.org/setup/pants
```
Eric-Arellano, line 44: What about adding shasum validation here? We do it for downloading Pex. cc @benjyw 

https://github.com/pantsbuild/setup/blob/dfc1a75baa01ee7948a4fd98bf4fd7f32df70797/pants#L45

Eric-Arellano (approved):

one_step_setup.sh
chrisjrn: It's not unreasonable. It _would_, however, mean we'd need to keep the one-step-setup script up to date whenever we  update the `pants` script. We'd catch that in the integration tests, but it'd be tedious.

Eric-Arellano: A bit tedious, but a small extra maintenance burden seems worth the better security. 

I cced @benjyw because we both have a lot of concern about supply chain attacks after reading https://www.amazon.com/This-They-Tell-World-Ends/dp/1635576059

benjyw: I'm not worried about the tedium, I am worried about the setup script failing for users because we forgot to update the sha, even if we eventually catch this in a test.

Eric-Arellano: How could that ever happen? The test would fail and block us from merging changes.

benjyw: The difference here is that we control both scripts, they even live in the same repo, so I'm not sure supply chain attack mitigation is a major concern here. What is the vector by which `pants` can get attacked but `setup` cannot?

Eric-Arellano: Yeah that's true, agreed it's less of a concern.

> What is the vector by which pants can get attacked but setup cannot?

Only one I can think of is the URL `https://static.pantsbuild.org/setup/pants` somehow is breached. Which..we right  now don't have any alerting to that happening. So maybe an error telling users to open a bug at this repo is a nice insurance policy.

And then beyond actual security concerns, there is perhaps a theatrical element to it of making the script look more secure. See https://github.com/pantsbuild/setup/pull/124#discussion_r887232886 for why we named the script `.sh`.

--

The only downside I can see is it's going to be more annoying to edit `./pants`, right? (Valid concern)

I don't think the concern about us forgetting to update the value can happen, thanks to CI.

chrisjrn: So as written, and as how our tests currently work, it will be _quite_ tedious to check that `one_step_setup.sh` actually works in a test. That's because the script fetches from `https://static.pantsbuild.org`, rather than the newly updated version of the script.

That means, we either need to spawn a web server in the tests (not insurmountable, just tedious) and hope that the integration test still succeeds when we stop using a false server (so the integration test is less trustworthy); or we need to do a two-step PR process when we update `./pants` (once to update `pants` and once to update `one_step_setup.sh`).

We may want to follow up on this one later, if you ask me.

Eric-Arellano: Ah okay, good point.

PR #127, opened 2022-08-30
Fix scrub msg in pants script and upgrade PEX version

dheerajramchandani:
I am opening this PR to modify the function `scrub_env_vars()` in the `pants` script so that it scrubs all `PEX_*` environment variables except `PEX_ROOT`. I have also upgraded the variables `_PEX_VERSION` and `_PEX_EXPECTED_SHA256` in the script but in a separate commit just in case we want to cherry pick and merge only the one related to scrubbing PEX_* environment variables.


jsirois (approved):
Thanks alot @dheerajramchandani! This looks good to me but I'll hold off merging until you give the green light on the unbound variable issue you're seeing.

dheerajramchandani:
Hi @jsirois, I'd prefer following fix rather adding an empty element:

`if [[ ! ${#pex_env_vars[@]} -eq 0 ]]; then`

because the empty entry gets assigned to the variable `pex_env_vars_to_scrub` and it still shows a `Scrubbing  ` (Scrubbing spaces) warning.

Please let me know what you think and I'd be happy to make a change.

pants
```diff
@@ -283,10 +283,12 @@ function scrub_env_vars {
   # Ensure the virtualenv PEX runs as shrink-wrapped.
   # See: https://github.com/pantsbuild/setup/issues/105
   local -r pex_env_vars=(${!PEX_@})
-  local -r pex_env_vars_to_scrub="${pex_env_vars[@]/PEX_ROOT}"
-  if [[ -n "${pex_env_vars_to_scrub[@]}" ]]; then
-    warn "Scrubbing ${pex_env_vars_to_scrub[@]}"
-    unset ${pex_env_vars_to_scrub[@]}
+  if [[ ! ${#pex_env_vars[@]} -eq 0 ]]; then
```
jsirois, line 286: What happened to the check I experimentally checked? Bash makes the sort of check you're doing here easier too with something like `if (( ${#pex_env_vars[@]} > 0 )); then`, but the single array declaration seems better. This extra check begs the question of why are we jumping though this extra check hoop?

dheerajramchandani: I provided an explanation here: https://github.com/pantsbuild/setup/pull/127#issuecomment-1232125314

jsirois: Alright, sounds good enough.

jsirois (approved):

PR #128, opened 2022-10-05
Add 'bootstrap tools' (bootstrap-cache-key) via PANTS_BOOTSTRAP_TOOLS=1

huonw:
This adds functionality that can introspect the bootstrapping process, similar to `PEX_TOOLS=1 ./some.pex ...` being able to introspect a PEX file.

The motivating tool, added here, is `PANTS_BOOTSTRAP_TOOLS=1 ./pants bootstrap-cache-key` that prints a somewhat opaque string designed to be used as a key for CI caching systems. This is designed to help with https://github.com/pantsbuild/actions/issues/5, and is prompted by the discussion in https://github.com/pantsbuild/actions/pull/6#discussion_r978745702.

For example, on my system:

```
$ PANTS_BOOTSTRAP_TOOLS=1 ./pants bootstrap-cache-key
os_name=Darwin arch=arm64 python_path=/Users/huon/.pyenv/shims/python3.9 python_version=Python 3.9.10 pex_version=2.1.103 virtualenv_requirements_sha256=80b5a45ee3ee507e268799305d4e5e347c7e8346df6551b6a83df05396b3d941 pants_version=2.13.0
```

(This PR is a quick sketch, please lemme know if it doesn't make sense!)

PR #129, opened 2022-10-05
Add version bootstrap tool subcommand, for capability checking

huonw:
This expands on #128 to add another bootstrap tool subcommand: `bootstrap-version`. This is designed to just be a simple number that's bumped whenever there's a feature scripts might need to query for. The version number is also automatically checked against the (numeric) value of the `PANTS_BOOTSTRAP_TOOLS` key.

For instance, suppose the bootstrap tools already existed (at version 1), and then `bootstrap-cache-key` command was added, the version number would be bumped to 2, and then a consumer (such as https://github.com/pantsbuild/actions/pull/6) could run something like:

```shell
PANTS_BOOTSTRAP_CACHE_KEY=$(PANTS_BOOTSTRAP_TOOLS=2 ./pants bootstrap-cache-key)
```

Notes:

- I don't personally have concrete uses cases for this, but potentially adding different output like https://github.com/pantsbuild/setup/pull/128#pullrequestreview-1131761160 might use it.
- The subcommand name is `bootstrap-version` rather than `version` to avoid conflict with the subcommand on real-`pants` (i.e. `./pants version` without `PANTS_BOOTSTRAP_TOOLS=1` works, and prints the underlying pants version like 2.13.0, whereas `./pants bootstrap-version` errors out with `Unknown goal: bootstrap-version`)
- It's unfortunate I didn't add this in #128, because there may be people who installed the script supporting `PANTS_BOOTSTRAP_TOOLS` but not supporting this versioning. Maybe the window is small enough that that doesn't matter?

### DIFF
--- a/pants
+++ b/pants
@@ -68,6 +68,8 @@ COLOR_GREEN="\x1b[32m"
 COLOR_YELLOW="\x1b[33m"
 COLOR_RESET="\x1b[0m"
 
+INSTALL_URL="https://www.pantsbuild.org/docs/installation"
+
 function log() {
   echo -e "$@" 1>&2
 }
@@ -144,7 +146,7 @@ function determine_pants_version {
   if [[ -z "${pants_version}" ]]; then
     die "Please explicitly specify the \`pants_version\` in your \`pants.toml\` under the \`[GLOBAL]\` scope.
 See https://pypi.org/project/pantsbuild.pants/#history for all released versions
-and https://www.pantsbuild.org/docs/installation for more instructions."
+and $INSTALL_URL for more instructions."
   fi
   pants_major_version="$(echo "${pants_version}" | cut -d '.' -f1)"
   pants_minor_version="$(echo "${pants_version}" | cut -d '.' -f2)"
@@ -393,6 +395,10 @@ function bootstrap_pants {
 
 function run_bootstrap_tools {
   # functionality for introspecting the bootstrapping process, without actually doing it
+  if [[ "${PANTS_BOOTSTRAP_TOOLS}" -gt "$SCRIPT_VERSION" ]]; then
+      die "$0 script (bootstrap version $SCRIPT_VERSION) is too old for this invocation (with PANTS_BOOTSTRAP_TOOLS=${PANTS_BOOTSTRAP_TOOLS}).
+Please update it by following $INSTALL_URL"
+  fi
 
   case "${1:-}" in
     bootstrap-cache-key)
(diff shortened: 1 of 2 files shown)


benjyw (approved):
I like it!

benjyw:
> * It's unfortunate I didn't add this in [Add 'bootstrap tools' (bootstrap-cache-key) via PANTS_BOOTSTRAP_TOOLS=1 #128](https://github.com/pantsbuild/setup/pull/128), because there may be people who installed the script supporting `PANTS_BOOTSTRAP_TOOLS` but not supporting this versioning. Maybe the window is small enough that that doesn't matter?

I think it's fine - window is small.

> * The checking could be packaged into the bootstrap script itself: the bootstrap script check that `SCRIPT_VERSION >= PANTS_BOOTSTRAP_TOOLS`, i.e. the example scenario above could be collapsed to just: `PANTS_BOOTSTRAP_CACHE_KEY=$(PANTS_BOOTSTRAP_TOOLS=2 ./pants bootstrap-cache-key)`. Thoughts?

That could be neat! So the script runs if it's later than PANTS_BOOTSTRAP_TOOLS and fails otherwise?

pants
```diff
@@ -412,17 +416,22 @@ function run_bootstrap_tools {
...
 Subcommands:
   bootstrap-cache-key
     print an opaque that can be used as a key for accurate and safe caching of the pants bootstrap directories
+  bootstrap-version
+    print a version number for the bootstrap script itself
 EOF
       ;;
     *)
-      die "Unknown subcommand for bootstrap tools (due to PANTS_BOOTSTRAP_TOOLS=1): $1 "
+      die "Unknown subcommand for bootstrap tools: $1. Do you mean to run without PANTS_BOOTSTRAP_TOOLS=1? Or, update this script (https://www.pantsbuild.org/docs/installation)?"
```
kaos: I would like a `bootstrap-update` command, that fetches the latest setup script and saves it over `$0`.

kaos (approved):

huonw:
> That could be neat! So the script runs if it's later than PANTS_BOOTSTRAP_TOOLS and fails otherwise?

Yeah, exactly. I've implemented something along those lines.

pants
huonw: That'd be cool! (Although I won't personally work on it, not in the short term, anyway)

kaos: https://github.com/pantsbuild/setup/issues/131

pants
```diff
@@ -140,7 +146,7 @@ function determine_pants_version {
   if [[ -z "${pants_version}" ]]; then
     die "Please explicitly specify the \`pants_version\` in your \`pants.toml\` under the \`[GLOBAL]\` scope.
 See https://pypi.org/project/pantsbuild.pants/#history for all released versions
-and https://www.pantsbuild.org/docs/installation for more instructions."
+and $INSTALL_URL for more instructions."
```
benjyw:
```suggestion
and ${INSTALL_URL} for more instructions."
```

pants
```diff
@@ -389,6 +395,10 @@ function bootstrap_pants {
 
 function run_bootstrap_tools {
   # functionality for introspecting the bootstrapping process, without actually doing it
+  if [[ "${PANTS_BOOTSTRAP_TOOLS}" -gt "$SCRIPT_VERSION" ]]; then
```
benjyw:
```suggestion
  if [[ "${PANTS_BOOTSTRAP_TOOLS}" -gt "${SCRIPT_VERSION}" ]]; then
```

pants
```diff
@@ -389,6 +395,10 @@ function bootstrap_pants {
...
+      die "$0 script (bootstrap version $SCRIPT_VERSION) is too old for this invocation (with PANTS_BOOTSTRAP_TOOLS=${PANTS_BOOTSTRAP_TOOLS}).
```
benjyw:
```suggestion
      die "$0 script (bootstrap version ${SCRIPT_VERSION}) is too old for this invocation (with PANTS_BOOTSTRAP_TOOLS=${PANTS_BOOTSTRAP_TOOLS}).
```

pants
```diff
@@ -389,6 +395,10 @@ function bootstrap_pants {
...
+Please update it by following $INSTALL_URL"
```
benjyw:
```suggestion
Please update it by following ${INSTALL_URL}"
```

pants
```diff
@@ -412,21 +422,38 @@ function run_bootstrap_tools {
       )
       echo "${parts[*]}"
       ;;
+    bootstrap-version)
+      echo "$SCRIPT_VERSION"
```
benjyw:
```suggestion
      echo "${SCRIPT_VERSION}"
```

benjyw (approved):
LGTM mod the best-practice of bracing vars.

huonw:
Oops, thanks for that.

PR #133, opened 2022-11-11
Don't `cd` into temp dirs when boostrapping Pants.

### LINKED ISSUE #132: `cd`s in bootstrapping logic prevent use of `asdf`-installed Python versions from `<project-root>/.tool-versions`
[`asdf`](https://asdf-vm.com/) is a CLI for managing multiple runtime versions of other tools (i.e. `python`). To use an `asdf`-managed tool, you:
1. `asdf install` it (i.e. `asdf install python 3.7.12`)
2. Write the installed version into a `.tool-versions` file

`.tool-versions` are hierarchical - for example if you have `~/.tool-versions`:
```
python 3.7.12
```
and also `~/project/.tool-versions`:
```
python 3.8.9
```
Then running `python --version` from within `~/project` will print `3.8.9` instead of `3.7.12`.

---

Pants' `python` backend has [dedicated support](https://www.pantsbuild.org/docs/python-interpreter-compatibility#changing-the-interpreter-search-path) for locating `asdf`-installed Pythons within both `~/.tool-versions` and in `<project-root>/.tool-versions`, but my team has found that the bootstrapping process only ever uses versions from `~/.tool-versions`. This happens because the bootstrapping process `cd`s into temp dirs at various points, preventing `asdf` from seeing the project-local `.tool-versions`.

I think we can refactor the script to avoid the `cd`s and stay within the project root, allowing local `asdf` to work.

danxmoran:
Closes #132 

Changing the work dir prevents tools like `asdf` from seeing config files in the project root, resulting in confusing errors / use of unexpected Python versions.

### DIFF
--- a/pants
+++ b/pants
@@ -265,14 +265,13 @@ function bootstrap_pex {
       mkdir -p "${PANTS_BOOTSTRAP}"
       local staging_dir
       staging_dir=$(tempdir "${PANTS_BOOTSTRAP}")
-      cd "${staging_dir}"
       curl --proto "=https" \
            --tlsv1.2 \
            --silent \
            --location \
-           --remote-name \
+           -o "${staging_dir}/pex" \
            "${_PEX_URL}"
-      fingerprint="$(compute_sha256 "${python}" "pex")"
+      fingerprint="$(compute_sha256 "${python}" "${staging_dir}/pex")"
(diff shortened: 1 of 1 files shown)


benjyw (approved):

jsirois (approved):

kaos (approved):

benjyw:
@danxmoran, as a Maintainer you should be able to merge now! Let me know if not.

PR #134, opened 2022-11-15
add support for overriding pants_version on the command line

tgolsson:
This would help support running different pants version easily on CI and while testing various versions, as opposed to editing the pants.toml file. Tested it locally on my plugin repos and it seems to work. Unfortunately I can't run the tests as I don't have a Python 3.7 on hand. They also don't seem to validate what is actually ran, but I added a case at least.

In my OCI backend repo I get this:

```console
$ PANTS_VERSION=2.14.0 pants --version
2.14.0

$ PANTS_VERSION=2.15.0a1 pants --version
20:51:45.74 [WARN] DEPRECATED: using `SubsystemRule(SkopeoTool)` is scheduled to be removed in version 2.17.0dev0.

Use `*SkopeoTool.rules()` instead.
20:51:45.74 [WARN] DEPRECATED: using `SubsystemRule(RuncTool)` is scheduled to be removed in version 2.17.0dev0.

Use `*RuncTool.rules()` instead.
20:51:45.74 [WARN] DEPRECATED: using `SubsystemRule(UmociTool)` is scheduled to be removed in version 2.17.0dev0.

Use `*UmociTool.rules()` instead.
2.15.0a1

$ pants --version
<snip warnings as above>
2.15.0a0
```

So all three report the expected version, and given the warnings for 2.15 alphas but not 2.14 it does switch the code too.

### DIFF
--- a/pants
+++ b/pants
@@ -26,6 +26,10 @@ fi
 #  locate the main branch SHA, set PANTS_SHA=<SHA> in the environment, and run this script as usual.
 #
 # E.g., PANTS_SHA=725fdaf504237190f6787dda3d72c39010a4c574 ./pants --version
+#
+# You can also use PANTS_VERSION=<VERSION> to override the config version that is in the pants.toml file.
+#
+# E.g., PANTS_VERSION=2.13.0 ./pants --version
 
 PYTHON_BIN_NAME="${PYTHON:-unspecified}"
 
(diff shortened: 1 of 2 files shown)


kaos (approved):
Tack så mycket! :)

tgolsson:
OK; test works too well because the pants.toml ends up being incompatible. I'll get 3.7 setup tomorrow and validate with a version that's compatible.

tgolsson:
Updated with test fixed.

PR #135, opened 2022-11-18
Update shim for new VERSION location

thejcannon:
See https://github.com/pantsbuild/pants/pull/17563 for the pants repo change.

### DIFF
--- a/pants
+++ b/pants
@@ -351,7 +351,7 @@ function get_version_for_sha {
                         --fail \
                         --silent \
                         --location \
-                        "https://raw.githubusercontent.com/pantsbuild/pants/${sha}/src/python/pants/VERSION")"
+                        "https://raw.githubusercontent.com/pantsbuild/pants/${sha}/src/python/pants/version/VERSION")"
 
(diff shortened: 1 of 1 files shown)


chrisjrn (changes requested):
Soooo, I think we might actually need to make the `PANTS_FROM_SHA` functionality a bit smarter, because it’s going to break if you’re testing commits a few months either side of now -- you need the new script for newer versions, and the old script for older versions.

Can we ask GitHub the timestamp of a specific Pants commit and then decide on the URL based on whether it's older or newer than a specific timestamp?

chrisjrn:
It looks like you can use the GitHub compare commits endpoint without authentication:

```
% curl \
  -H "Accept: application/vnd.github+json" \
  https://api.github.com/repos/pantsbuild/pants/compare/169d2e613cad7b1bd2593ae9e747100458bf2e5f...66edd8917f5c57b8ab0e36c88dc556142b321732 | grep \"status\"
  "status": "behind",
```

That `status` field will give either `ahead`, `behind`, or `identical`. So you can check to see if the requested SHA is `behind` the changeover commit, and in that case, use the old URL version.

thejcannon:
Going to drop in favor of a symlink

PR #137, opened 2023-01-15
A pantsup.sh script to easily install the latest scie-pants.

benjyw:
Once merged, you just

```
curl -sSLf https://static.pantsbuild.org/setup/pantsup.sh | sh
```

And the right scie-pants for your platform, at the latest version, gets installed to ~/bin/pants.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -30,14 +30,14 @@ deps =
     isort
     black
 commands =
-    isort --recursive tests/ --apply
+    isort tests/ --apply
     black .
 
 [testenv:format-check]
 deps =
    {[testenv:format-run]deps}
 commands =
-    isort --recursive tests/ --check-only
+    isort tests/ --check-only
     black . --check
 
 [testenv:lint]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,13 +19,15 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Setup Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ env.PYTHON_VERSION }}
+      - name: Shellcheck
+        run: shellcheck pantsup.sh
       - name: Run Checks
-        uses: pantsbuild/actions/run-tox@95209b287c817c78a765962d40ac6cea790fc511
+        uses: pantsbuild/actions/run-tox@v2
         with:
           tox-env: format-check,lint,typecheck
   unit-tests:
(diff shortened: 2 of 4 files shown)


tox.ini
```diff
@@ -30,14 +30,14 @@ deps =
     isort
     black
 commands =
-    isort --recursive tests/ --apply
+    isort tests/ --apply
```
benjyw, line 33: isort 5.0.0 recurses always, and this flag is deprecated with a loud warning

jsirois:
I had done this already, but it's here: https://github.com/pantsbuild/scie-pants/blob/main/get-pants.sh

benjyw:
Ah dammit, I had missed that completely.

jsirois:
I think you tweeted it at someone! You're as bad as me!

benjyw:
My brain is like the inside of a chicken nugget at this point.

benjyw:
I would like to have this to the `pantsbuild/setup` repo though (I can take your get-pants.sh, or merge the two), because we can serve that via static.pantsbuild.org, for which we have analytics.

jsirois:
Ok... I had not looked into DNS but did not think that would be hard. I had looked into GH pages a bit. Here's my tracking issue: https://github.com/pantsbuild/scie-pants/issues/35

benjyw:
We already have all this set up for static.pantsbuild.org (including Google Analytics) so seems easiest to get started that way.

jsirois:
Ok, how about you take over https://github.com/pantsbuild/scie-pants/issues/35 and deal with relocation / removal of IT and what have you. I'm happy to hand all that off if you're on a war path.

benjyw:
If it's more convenient for the script to live here, I can also create a new domain for it etc.

jsirois:
Yeah, as you see fit. The only oddness is, in the long run the scie-pants will live in scie-pants and the script to install it will live here along with nothing else. The only inconvenience is the IT is split away from scie-pants and will live here with, again in the long run, IT infra already existing back in scie-pants. I don't think any of these considerations are actually hugely important either way and action trumps opining; so I say charge.

benjyw:
True, once we get rid of the old `pants` and `one_step_setup.sh` scripts, this whole repo will become pretty unnecessary, and we can move this script back into scie-pants, and move the pretty URL (or create a prettier one) to point there, with analytics etc. 

But for now as you say, I want to just move things forward. And all that will be opaque to end users anyway.

benjyw:
OK, so the commits are now useful to review independently. The first is just copying the existing get-pants.sh over, and renaming it pantsup.sh because puns.

Then subsequent commits tweak that script slightly, to add terminal colors, to tweak the log messages, and to avoid placing the "scie-pants" terminology front and center, since this is for users just getting started, and I'd like to shield them from lotsa names of things.

Also, I deleted the bit at the end where this script invokes `pants` itself. That seemed unnecessary and potentially confusing if you ran this not in a repo. Instead there's a message telling you how `pants` works.

jsirois (approved):
As long as you don't curse yourself for having to explain `pantsup.sh` is - despite the confusing name - _not_ what you use to update the Pants launcher or the Pants version, its just what you use to install the Pants lanucher when you don't already have it, looks good.

benjyw:
You could use it to update the pants launcher, no? If I rerun it it'll download the latest and overwrite the old one?

kaos (approved):

jsirois:
> You could use it to update the pants launcher, no? If I rerun it it'll download the latest and overwrite the old one?

You could, but that is definitely not the intent! I built a whole self-update thing into the scie-pants binary with an issue to track self-update nagging and all of that. For the continual comparison point, that's not how rustup works - or at least its not how I use it! I install once via curl then use its self-update mechanism.

PR #138, opened 2023-01-25
add shfmt ci check

asherf:


.github/workflows/ci.yml
```diff
@@ -26,6 +26,10 @@ jobs:
           python-version: ${{ env.PYTHON_VERSION }}
       - name: Shellcheck
         run: shellcheck pantsup.sh
+      - uses: luizm/action-sh-checker@v0.5.0
```
benjyw, line 29: This runs on every `.sh` file?

benjyw (approved):

.github/workflows/ci.yml
asherf: yes. will add some PR to address the issues detected by this check.
see: https://github.com/pantsbuild/setup/pull/139 (first one), will have more.
so this PR can pass CI checks.

benjyw (commented):
Looking at the results in the other PRs - the tab indents are crushing my soul. We could run with `-i 4` which will use 4-space indents, which I think all would greatly prefer. For one thing, everyone's editors are set up to emit spaces when you hit tab, so actually getting those tabs right would be impossible without running the tool.

benjyw:
I have a pending change to use Pants in this repo, which will give us shfmt, shellcheck and all the rest for free. So that might be a better path than manually adding this sort of thing.

benjyw:
Closing this due to https://github.com/pantsbuild/setup/pull/142

PR #141, opened 2023-01-31
Rename pantsup.sh to get-pants.sh.

benjyw:
The original name was a bit of pun indulgence on my part. But now that we're suggesting that people check the script into the repo, it makes sense to rename it, for two reasons:

1) get-pants.sh is more descriptive, so a user rummaging around
   in the repo for "how do I install this Pants thing" will
   have more obvious clues.
2) To avoid annoying tab-completion competition with the `pants` launcher
   binary, or the `./pants` script if it exists.

We symlink pantsup.sh to get-pants.sh for now, as some users may already be relying on the old name. We can consider how long to continue to do that once we change the documentation to point to get-pants.sh.


chrisjrn:
I am definitely in favour of not having `pantsup` or anything else to do with the position of pants be in the name for the script.

jsirois (approved):

benjyw:
PS I confirmed before making this change that Github Pages serves symlinks as expected, and have now verified that both of these work as expected:

curl --proto '=https' --tlsv1.2 -fsSL https://static.pantsbuild.org/setup/pantsup.sh
curl --proto '=https' --tlsv1.2 -fsSL https://static.pantsbuild.org/setup/get-pants.sh

PR #143, opened 2023-06-17
get-pants.sh: Fix deletion of temp files

wlritchi:
get-pants.sh includes a simple `gc` helper function to mark temporary files and delete them when the script exits. The script mistakenly passed it the string `workdir` rather than the value of the workdir variable. This commit changes `workdir` to `"${workdir}"`.


benjyw (approved):
Ooooooof, that's nasty. Thanks for the fix!

PR #144, opened 2023-11-07
Install scie-pants in `~/.local/bin` instead of `~/bin`

cognifloyd:
Using `~/bin` does not follow the [XDG based directory spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) which recommends `~/.local/bin` and does not offer an `XDG_` env var alternative similar to other paths.

I first mentioned doing this PR in
https://github.com/pantsbuild/actions/pull/21#discussion_r1185651144

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,9 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+      - name: Ensure default install path is on PATH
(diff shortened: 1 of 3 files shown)


kaos (approved):

cognifloyd:
Oh joy. looks like I have to change the init action too because not some runner versions are missing `~/.local/bin` on `PATH`. Or I have to add the path in this CI workfdlow...

